PR #63, opened 2015-01-29
Fix $checkbox is undefined when trying to select aggregates that don't exist

### LINKED ISSUE #62: $checkbox is undefined bootstrap-multiselect.js:729:20 when aggregates are selected that don't exist for a probe
STR:

1. Load a probe that allows you to choose only Mean/Median/etc. (not just the submission count) e.g. PWMGR_NUM_SAVED_PASSWORDS.
2. Uncheck some of the checkboxes of data to see on the graph e.g. only leave submissions, mean and median. The url parameter updates to include `&aggregates=Submissions!Mean!median`
3. Switch to a probe that doesn't let you graph Mean/Median e.g. A11Y_CONSUMERS

Expected result:
The URL should update to indicate the selected probe name (e.g. A11Y_CONSUMERS) and the histogram should update.

Actual result:
A JS error occurs: TypeError: $checkbox is undefined bootstrap-multiselect.js:729:20
The URL doesn't update.
The histogram doesn't update (still shows the data from the last probe).

Showing the wrong histogram for a probe is a big problem if decisions are getting made based on the data.

mnoorenberghe:
Fixes #62.

r? @jonasfj

### DIFF
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -422,6 +422,11 @@ function setAggregateSelectorOptions(options, changeCb, defaultToAll) {
   var prevOptions = [];
   $("#aggregateSelector option").each(function() { prevOptions.push($(this).val()); });
   var prevSelected = $("#aggregateSelector").multiselect("getSelected").val() || [];
+  // Find the intersection of the previously selected and the valid options so
+  // we don't get an error trying to select non-existent options for a probe.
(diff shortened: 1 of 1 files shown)


jonasfj:
Wow, what a lovely corner case; thanks for the fix :)

PR #67, opened 2015-04-01
Check for the HTTPS version of releaseURL too

fmarier:
https://bugzilla.mozilla.org/show_bug.cgi?id=1119876

I tested this change manually by dropping this file into my local webserver and hacking my `/etc/hosts` to point `telemetry.mozilla.org` to `127.0.0.1`.


jonasfj:
This good to me... It'll only have effect if included with `http://`... besides it'll only ever print a warning, so not going to break anything :)

vdjeric:
Mark: please comment in bug 1119876 when this gets deployed

mreid-moz:
Commented/resolved bug 1119876.

PR #68, opened 2015-05-05
Various UI Fixes

Uberi:
* Fix the main screen notice of selection bias.
* Make the filtering system more flexible:
    * Add support for "fixed options", which are always implicitly included in the filter, and use this to implement `save_session` filtering.
  * Add support for labeling options with values that are not the same as their actual values, and use this to implement friendly OS/version names. For example, "WINNT" is now shown as "Windows", and Darwin "10.0.0" is shown as "10.0.0 (Snow Leopard)".
  * Remove the "Any App" option since it is not very useful (previously known as "appName*").

### DIFF
--- a/src/jquery.telemetry.js
+++ b/src/jquery.telemetry.js
@@ -301,12 +301,14 @@ $.widget("telemetry.histogramfilter", {
 
       // Get histogram
       histogram = filter.histogram;
-
+      
       // Get selected option
-      var option = filter.select.val();
+      var option = (typeof filter.select.val === "function") ?
+                   filter.select.val() : filter.select;
+      option = this._getOptionRealValue(histogram.filterName(), option);
 
       // If selected option isn't the default option and the option is available
-      if (option != histogram.filterName() + "*" &&
+      if (option != this._getStarName(histogram.filterName()) &&
           histogram.filterOptions().indexOf(option) !== -1) {
         
         // Filter the histogram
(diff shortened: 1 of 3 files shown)


Uberi:
* Fixed mappings for OS X versions.
* Added "Any Build" and names for architectures.
* Improve mapping lookup and make it a bit cleaner.
* Squashed commits.

src/jquery.telemetry.js
```diff
@@ -318,6 +320,90 @@ $.widget("telemetry.histogramfilter", {
     return histogram;
   },
 
+  _getStarName: function histogramfilter__getStarName(filterName) {
+    var prettyNames = {"appName": "Any App", "osVersion": "Any Version", "arch": "Any Build"};
+    return prettyNames.hasOwnProperty(filterName) ? prettyNames[filterName] : "Any " + filterName;
+  },
+  
+  _windowsVersionNames: {"5.0": "2000", "5.1": "XP", "5.2": "XP Pro x64", "6.0": "Vista", "6.1": "7", "6.2": "8", "6.3": "8.1", "10.0": "10"},
+  _getOptions: function histogramfilter__getOptions(filterName, options) {
```
vdjeric: Rename _getOptions to _getHumanReadableOptions in a future pull request

src/jquery.telemetry.js
```diff
@@ -318,6 +320,90 @@ $.widget("telemetry.histogramfilter", {
...
+        return options;
+      }
+      
+      if (system === "WINNT") {
+        var versionNames = this._windowsVersionNames;
+        return options.map(function(option) {
+          return versionNames.hasOwnProperty(option) ? versionNames[option] : option;
+        });
+      } else if (system === "Darwin") {
+        var versionPrefixNames = {
```
vdjeric: Make these constants a top level member variable as well

src/jquery.telemetry.js
```diff
@@ -318,6 +320,90 @@ $.widget("telemetry.histogramfilter", {
...
+          for (var prefix in versionPrefixNames) {
+            if (option.startsWith(prefix)) {
+              return option + " (" + versionPrefixNames[prefix] + ")";
+            }
+          }
+          return option;
+        });
+      }
+    } else if (filterName === "arch") {
+      var archNames = {"x86": "32-bit", "x86-64": "64-bit"};
```
vdjeric: This too should be a member variable

src/jquery.telemetry.js
```diff
@@ -318,6 +320,90 @@ $.widget("telemetry.histogramfilter", {
...
+        if (this._windowsVersionNames[key] === value)
+          return key;
+      }
+      
+      // check for Darwin/OS X names
+      if (value.indexOf(" (") >= 0 && value[value.length - 1] === ")") {
+        return value.split(" ")[0];
+      }
+    } else if (filterName == "arch") {
+      var realArchNames = {"32-bit": "x86", "64-bit": "x86-64"};
```
vdjeric: this should be determined by iterating over the archNames member variable

src/jquery.telemetry.js
```diff
@@ -318,6 +320,90 @@ $.widget("telemetry.histogramfilter", {
...
+      return options.map(function(option) {
+        return archNames.hasOwnProperty(option) ? archNames[option] : option;
+      });
+    }
+    return options;
+  },
+  
+  _getOptionRealValue: function histogramfilter__getOptionRealValue(filterName, value) {
+    if (filterName == "OS") {
+      var realSystemNames = {"Windows": "WINNT", "OS X": "Darwin"};
```
vdjeric: Just put all the mappings at the top level of this object

PR #70, opened 2015-05-07
Fix mixed-content errors when accessing some dashboards using https.

mreid-moz:
Fixes [bug 1162265](https://bugzilla.mozilla.org/show_bug.cgi?id=1162265).

### DIFF
--- a/chromehangs/index.html
+++ b/chromehangs/index.html
@@ -9,9 +9,9 @@
     <script type="text/javascript" src="https://code.jquery.com/jquery.js"></script>
     <script type="text/javascript" src="https://code.jquery.com/ui/1.10.4/jquery-ui.js"></script>
     <script type="text/javascript" src="chromehangs.js"></script>
-    <script src="http://jquery-csv.googlecode.com/git/src/jquery.csv.js"></script>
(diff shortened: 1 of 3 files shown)


chromehangs/index.html
```diff
@@ -9,9 +9,9 @@
     <script type="text/javascript" src="https://code.jquery.com/jquery.js"></script>
     <script type="text/javascript" src="https://code.jquery.com/ui/1.10.4/jquery-ui.js"></script>
     <script type="text/javascript" src="chromehangs.js"></script>
-    <script src="http://jquery-csv.googlecode.com/git/src/jquery.csv.js"></script>
+    <script src="//jquery-csv.googlecode.com/git/src/jquery.csv.js"></script>
```
vdjeric: what is the src="//<site name>/.." format? it defaults to http or https based on how the top-level page was loaded? why not always use https?

mreid-moz: Yes, it just matches the top-level URI scheme.  It could just use `https` all the time if you prefer. Using `//` keeps the current behaviour if you access telemetry.m.o via http, while fixing the https case.

PR #71, opened 2015-05-08
Replace d3.js with Chart.js and 

Uberi:
* Chartjs results in higher-quality graphs.
    * That means no more overlapping labels in histograms!
    * Multi-tooltips are essentially legends that show up on hovering over the graph - the legend panel on the right is made unnecessary and is removed.
    * Added explicit date selector for graphs.
    * Histogram tooltips are positioned better, right over the bars they correspond to.
* Lighter page overall.
    * Removed d3 entirely.
    * Our custom d3-based graphing addons are no longer needed, which makes the codebase a lot smaller.
* Refactored the updating code and made style a bit more consistent.

The commit history is a little messy right now, should I squash them?


index.html
```diff
@@ -4,13 +4,23 @@
...
-    <link rel="stylesheet"
-          href="//netdna.bootstrapcdn.com/font-awesome/4.0.2/css/font-awesome.min.css">
-    <link rel="stylesheet" href="lib/nv.d3.css">
+    
+    <script src="//code.jquery.com/jquery-1.10.1.min.js"></script>
+    
+    <script src="//netdna.bootstrapcdn.com/bootstrap/3.0.2/js/bootstrap.min.js"></script>
+    <link rel="stylesheet" href="//netdna.bootstrapcdn.com/bootstrap/3.0.2/css/bootstrap.min.css">
+    <script src="lib/bootstrap-select.min.js"></script>
     <link rel="stylesheet" href="lib/bootstrap-select.min.css">
```
vdjeric: why are some libs copied locally and some referenced remotely?

src/dashboard.js
```diff
@@ -610,8 +594,24 @@ function fmt(number) {
   if (number == Infinity) return "Infinity";
   if (number == -Infinity) return "-Infinity";
   if (isNaN(number)) return "NaN";
-  var prefix = d3.formatPrefix(number, 's');
-  return Math.round(prefix.scale(number) * 100) / 100 + prefix.symbol;
+  var mag = Math.abs(number);
```
vdjeric: you should write a formatPrefix function, with fewer else-if's :)

src/dashboard.js
```diff
@@ -780,68 +780,189 @@ function renderHistogramGraph(hgram) {
...
+function updateRendering(hgramEvo, lines) {
+  // Filter dates
+  var dates = hgramEvo.dates();
+  var start = dates[0], end = dates[dates.length - 1];
+
+  // Update the start and end range and update the selection if necessary
+  var picker = $("#dateRange").data("daterangepicker");
+  var startMoment = moment(start), endMoment = moment(end);
+  picker.setOptions({
+    format: "MM/DD/YYYY h:mm A",
```
vdjeric: I don't think we need time for the picker.. maybe just a tooltip to explain that the range is inclusive
In particular, the Nightly with buildID "May 8" is generated shortly after midnight of May 9th based on Eastern time.. so let's not mention times

vdjeric:
Leave the commit-history as is, it's fine. Btw, we don't have a good person to review your dashboard patches, Mark and Jonas are busy with other things and they haven't done dashboard development in a long time. I'm not familiar with the codebase either and I don't have time to ramp up on it. So you're a little bit on your own with these patches :) 
So instead, I would like you to prepare a simple (manual) test plan for you and future devs to follow when testing new patches. Write up the test plan in Google Docs using your mozilla.com account

vdjeric:
Add series looks broken when "evolution over buildID" is selecte: http://bit.ly/1FUEJxM
Steps to reproduce: 
1. Display only median for GC_MS for Nightly 40
2. Choose "evolution over buildID"
2. Add Nightly 39 and Nightly 38

I think it's because you don't expand the X-axis range when new series are added

Uberi:
Updated:

* Use CDN for bootstrap-multiselect (the remaining local libraries are not available via CDN).
* Merge changes in from master.
* `fmt(NUM)` is more flexible.
* No more times in the date range picker (assume midnight as start and end of day).
* Fix bug where only first series is used to calculate chart bounds.
* Fix bar colors.
* Fix CSV export.
* Fix tooltip labels not showing the formatted quantity.
* Fix tooltips not showing which series the point belongs to.

mreid-moz:
Thanks for including a live link - that is extremely helpful for review purposes.  If possible, please provide a live link for each PR.

vdjeric:
"Add series" needs more testing. I looked at median SIMPLE_MEASURES_FIRSTPAINT on Nightly 40 on Windows, then I added a second series for OS X and then added a third one for all OSes, the OS filter for the 3rd series is blank instread of "Any OS": http://bit.ly/1RGfy6W

vdjeric:
Additionally, when I hover over points in different series (which are identical except for OS), the tooltip doesn't tell me which line is for which OS and there's no way to find out: http://bit.ly/1bQFIm4

vdjeric:
And when I switch that last link http://bit.ly/1bQFIm4 from "Build dates" to "Calendar dates", I lose one of the series and two of the OS selectors are empty

Uberi:
These issues were fixed previously in another branch (`ui-fixes-2`), so I've merged it now - the issues no longer show up in those test cases or in various other series add/remove tests.

Also changed:

* New version of the charting library (which fixes a few small dataset and drawing issues).
* Tooltips now display the entire filter state, in forms like "nightly 40, Firefox 32-bit (Windows 7)".

All changes are still live at the link above.

vdjeric:
Does Google Charts support having two Y axes?

vdjeric:
There's some quirk with how the OS version selector interacts with cookies from the previous version. I was trying out http://anthony-zhang.me/telemetry-dashboard/ and sometimes I got a selector for "Windows 6.1" and sometimes I got a selector for "Windows 7"

https://www.dropbox.com/s/xld8jenox7eg369/Screenshot%202015-05-14%2017.52.34.png?dl=0

I wonder if there's any simple way to fix this.. Maybe add a version number to the cookies or do an extra check when interpreting the cookie data?


How to reproduce:
1. Clear cookies
2. Open http://anthony-zhang.me/telemetry-dashboard
3. Select SIMPLE_MEASURES_FIRSTPAINT

vdjeric:
Some of the plots Google Charts produces have sharp spikes. Any way we can turn that off?

Old: https://www.dropbox.com/s/l5ux6s2890ii940/Screenshot%202015-05-14%2018.07.30.png?dl=0
New: https://www.dropbox.com/s/q0w5ctxce1uded9/Screenshot%202015-05-14%2018.06.31.png?dl=0

vdjeric:
Unrelated to this pull request, there is some issue with Nightly 37 GC_MS Telemetry in the old & new dash, can you investigate?

vdjeric:
Also unrelated to the new dash, but does it seem like the "Ignore data points with very small number of submissions" feature isn't working properly? 

http://bit.ly/1K8YunW

vdjeric:
Also, can you create a fork with the old (current) dashboard and have it hosted at some other telemetry.mozilla.org URL? I'd like to switch to your new dash ASAP and keep the old one around as a backup for a while

vdjeric:
Another new enhancement request: What do you think about only showing the median series by default? If " Ignore data points with very small numbers of submissions" is checked by default and is working as it should, the median view alone should be enough for most people

Uberi:
1. Chartjs supports multiple axes through extending the scatter plot chart type - we'll probably need this for the Submissions line eventually.
2. I couldn't reproduce this issue by putting up 717cc9b9ea8383253ebe23ed2a7dd573b52ef942 on localhost, then swapping it for this branch, nor for the current `master`. Perhaps that was before the `ui-fixes-2` branch was merged? I pushed those changes live a little while after it was merged.
3. I've disabled Bezier curves in the graphs, so the spikes will no longer appear. Incidentally, drawing the evolution chart is noticeably faster now.
4. Seems like someone submitted a histogram with the date in 2414, so the X axis is blown totally out of scale. Should I add a filter that rejects dates far into the future or past? Is this something that should be done in the frontend?
5. I tested http://bit.ly/1K8YunW against http://mzl.la/1EL7Iim, and could not find any discernible differences (except the spike in the Bezier curve, which is a separate issue that is now fixed).
6. I've changed it so that if the median is available when the selector is updated, and the aggregates are not explicitly specified, then the median becomes the only selected option.

vdjeric:
- Don't plot points that don't have enough submissions: http://bit.ly/1KiIL5z
- I think there is an old bug that pops up when the padlock icon on top left is clicked: when the padlock is "unlocked", we show the distribution histogram even for multiple series and the page URL doesn't get updated as new series are added

vdjeric:
Merging, but can you fix the 2 issues I pointed out in another pull request?

Uberi:
These fixes, and a few others, should now be available in #74.

PR #72, opened 2015-05-11
UI fixes part 2.

Uberi:
* Implement all feedback from #68.
    * Top level mappings, rename functions, etc.
* Fix state loading from hash (previously the dropdowns would be blank).
* Sort Windows versions by release date, add in Windows 10 tech preview and Windows 95.

### DIFF
--- a/src/jquery.telemetry.js
+++ b/src/jquery.telemetry.js
@@ -305,7 +305,7 @@ $.widget("telemetry.histogramfilter", {
       // Get selected option
       var option = (typeof filter.select.val === "function") ?
                    filter.select.val() : filter.select;
-      option = this._getOptionRealValue(histogram.filterName(), option);
+      option = this._getHumanReadableOptionRealValue(histogram.filterName(), option);
 
       // If selected option isn't the default option and the option is available
       if (option != this._getStarName(histogram.filterName()) &&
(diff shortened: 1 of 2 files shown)


Uberi:
Now live for testing at http://fireman-ribbon-83064.bitballoon.com/.

vdjeric:
Why does the test site have the "reason" selector?

src/jquery.telemetry.js
```diff
@@ -325,52 +325,59 @@ $.widget("telemetry.histogramfilter", {
...
+      // get the currently selected OS
+      var system = $("#histogram-filters input.filter-OS").val() || this._getStarName("OS");
+      system = this._getHumanReadableOptionRealValue("OS", system);
       
       if (system === "WINNT") {
         var versionNames = this._windowsVersionNames;
-        return options.map(function(option) {
+        var versionOrder = this._windowsVersionOrder;
+        var starName = this._getStarName(filterName);
+        return options.sort(function(a, b) {
```
vdjeric: how often is this sort going to be done? is it only when the OS filter option is changed?

Uberi: Right, `_getHumanReadableOptions` is called whenever a select needs to be changed, and this particular line is only executed when the OS selector is set to "Windows".

PR #73, opened 2015-05-12
Use Web Workers for background processing

Uberi:
* Web Workers allow us to do the heavy lifting off of the UI thread, which keeps the window responsive.
* There is still a little bit of stutter when the charts are being drawn, which is unavoidable AFAIK.
* This adds a file to the Telemetry distribution called `telemetry-worker.js`, the background worker, as well as functions like `Telemetry.doAsync` for callback-style async worker operations.
* The worker is only used for histogram precomputation and entry counting, since those were the only real hotspots.
* A small optimization to bucketing also caused a gigantic speedup in histogram evolution creation.

### DIFF
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -54,7 +54,7 @@ function prepareData(state, hgramEvo) {
   var maxSubmissions = 0;
   // Whether we actually filter submissions is controllable via the
   // 'sanitize-pref' preference.
-  var sanitizeData = $('input[name=sanitize-pref]:checkbox').is(':checked');
+  var sanitizeData = $('#sanitize-pref').is(':checked');
   var pgState = getPageState();
   dataKey = state + " " + pgState.evoOver + " " +  pgState.sanitize;
   if (dataKey in cachedData) {
@@ -156,7 +156,7 @@ function getPageState() {
   pageState.aggregates = pageState.aggregates.filter(function(e) { return e !== "";});
   pageState.evoOver = $('input[name=evo-type]:radio:checked').val();
   pageState.locked = gSyncWithFirst;
-  pageState.sanitize = $('input[name=sanitize-pref]:checkbox').is(':checked');
+  pageState.sanitize = $('#sanitize-pref').is(':checked');
   pageState.renderhistogram = $('input[name=render-type]:radio:checked').val();
   return pageState;
 }
@@ -197,7 +197,7 @@ function restoreFromPageState(newPageState, curPageState) {
   }
 
   if (newPageState.sanitize !== undefined) {
-    $('input[name=sanitize-pref]:checkbox').prop('checked', toBoolean(newPageState.sanitize));
+    $('#sanitize-pref').prop('checked', toBoolean(newPageState.sanitize));
   }
 
   if (newPageState.evoOver !== undefined) {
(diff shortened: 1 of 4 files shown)


Uberi:
This is now live at http://performer-chimpanzee-56033.bitballoon.com/ for testing purposes.

v1/telemetry-worker.js
```diff
@@ -0,0 +1,62 @@
+importScripts("telemetry.js", "big.js")
+
+var gActions = {
+  "Histogram-count": function() {
```
vdjeric: use the same style as the other files, we don't quote function names or use hyphens inside the names

v1/telemetry.js
```diff
@@ -1468,8 +1514,8 @@ Histogram.prototype.geometricStandardDeviation = function() {
     );
   }
   var count       = this.count();
-  var log_sum     = _aggregate(DataOffsets.LOG_SUM, this);
-  var log_sum_sq  = _aggregate(DataOffsets.LOG_SUM_SQ, this);
+  var log_sum     = this.precomputeAggregateQuantity(Telemetry.DataOffsets.LOG_SUM);
+  var log_sum_sq  = this.precomputeAggregateQuantity(Telemetry.DataOffsets.LOG_SUM_SQ);
```
vdjeric: what is log sum squares used for?

Uberi: We use this to calculate the geometric means, as part of `Telemetry.Histogram.geometricMean()`.

Uberi:
Merged `chart-library` branch, which had various additional fixes and tweaks from the ui-fixes-2 branch.

The new version is now live at http://police-heel-16682.bitballoon.com/ (different from previous site).

vdjeric:
The filter options for the default histogram aren't loaded :) I'm going to merge your chart.js pull request, can you rebase this pull on top of it?

Uberi:
Ah, that was already fixed in the chart-library branch (I had to clear cookies first, however).

I've merged all the changes to master, and this branch is now live at http://anthony-zhang.me/telemetry-dashboard/.

src/dashboard.js
```diff
@@ -735,42 +737,43 @@ function renderHistogramGraph(hgram) {
...
+      barDatasetSpacing : 0,
+      barShowStroke: false,
+      scaleLabel: function(valuesObject) { return fmt(valuesObject.value); },
+      tooltipFontSize: 10,
+      tooltipTemplate: function(valuesObject) { return tooltipLabels[valuesObject.label] || valuesObject.label; },
+    });
+    
+    // Assign random colors to make it easy to differentiate between bars
+    gCurrentHistogramPlot.datasets[0].bars.forEach(function(bar) {
+      bar.fillColor = "hsla(" + Math.floor(Math.random() * 256) + ", 80%, 70%, 0.8)";
```
vdjeric: It would be nice to always use the same colors instead of random every time

v1/telemetry-worker.js
```diff
@@ -0,0 +1,62 @@
+importScripts("telemetry.js", "big.js")
+
+var gActions = {
+  Histogram_count: function() {
```
vdjeric: Just out of curiosity, have you time how much main-thread time is saved from moving these operations to a worker? if not, could you add some console logging code to time these operations and let me know what kinds of numbers you see

vdjeric:
Looks good, merging it. Let me know the timings + fix the random colors enhancements in another pull request

PR #74, opened 2015-05-20
Small chart fixes

Uberi:
* Don't draw points at all where there aren't enough submissions.
* Update the URL hash properly when filters are added while the locking is off.
* Make newly added filters duplicate the last filter rather than the first one.
* Fix a bug in which there would not be any filters for a series that was added at first.

### DIFF
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -299,10 +299,6 @@ function updateUrlHashIfNeeded() {
     return;
   }
 
-  if (anyHsLoading()) {
-    return;
-  }
-
   var pageState = getPageState();
   if (window.location.hash.split("#")[1] === pageStateToUrlHash(pageState)) {
     return;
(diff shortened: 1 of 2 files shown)


src/dashboard.js
```diff
@@ -457,10 +446,10 @@ Telemetry.init(function () {
   
   // Add series button
   $("#addVersionButton").click(function () {
-    var state = null;
+    var state = "nightly/40/SIMPLE_MEASURES_FIRSTPAINT/saved_session/Firefox";
```
vdjeric: what is this hardcoded string?

Uberi: It's the state that the first filter added to the dashboard will have - it's technically not allowed to happen in the UI, but if the cookie or URL hash is malformed or from an outdated version then this ensures that we don't simply stop showing filters like before.

vdjeric: ok

PR #75, opened 2015-05-21
Add old version of dashboard at /old

Uberi:
The old version of the dashboard needs to be made available on `telemetry.mozilla.org`.

The simplest way to do this is by adding it to the repository - the page needed a few changes to fix the broken links to other dashboards.


vdjeric:
Wish the deployment script could just let us create a branch from an earlier revision :(

PR #77, opened 2015-05-21
Fix future date filtering.

Uberi:
71d6fbe8751fed50d329df8d5ad575135eaefb6e was rebased from chart-fixes 2, which mixed up the merge a bit and resulted in an error.

### DIFF
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -936,16 +936,16 @@ function update(hgramEvos) {
     var futureCutoff = moment().add(1, "years").valueOf();
     series = $.map(series, function(entry, i) {
       // Update the bounds properly
-      if (point.x >= futureCutoff) {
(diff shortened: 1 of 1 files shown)


src/dashboard.js
```diff
@@ -936,16 +936,16 @@ function update(hgramEvos) {
...
-        continue;
-      }
       entry.values.forEach(function(point) {
-        if (start === null || point.x < start) {
-          start = point.x;
-        }
-        if (end === null || point.x > end) {
-          end = point.x;
+        if (point.x >= futureCutoff) {
+          console.log("Bad point; timestamp is far into the future: " + point.x);
```
vdjeric: Can you log the series info (channel, version, hgram name, filters) as well?

PR #78, opened 2015-05-22
Fix display when the timescale is small

Uberi:
When the timescale only shows one point, the scatter plot doesn't know what scale to use, and ends up being undefined.

We can specify the scale by adding an endpoint series to the plot.

This is now live at http://anthony-zhang.me/telemetry-dashboard/.

### DIFF
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -773,7 +773,7 @@ function renderHistogramEvolution(lines, minDate, maxDate) {
   var drawnLines = lines.filter(function(line) { return !line.disabled; });
   
   // Filter out the points that are outside of the time range
-  var minX = maxDate, maxX = minDate;
+  var minX = maxDate, maxX = minDate, minY = Infinity, maxY = -Infinity;
   var filteredDatasets = drawnLines.map(function (line) {
     return {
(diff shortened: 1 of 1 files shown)


src/dashboard.js
```diff
@@ -773,13 +773,28 @@ function renderHistogramEvolution(lines, minDate, maxDate) {
...
+        if (point.x > maxX) { maxX = point.x; }
+        return point;
+      }),
     };
-  })
+  });
+  
+  // Add a fake series to expand the bounds a bit, which makes the chart look nicer when the timescale is small (within a day or so)
+  filteredDatasets.push({
+      data: [{x: minX - 1000, y: 2798}, {x: maxX + 1000, y: 2798}],
```
vdjeric: what are the implications of hardcoding 2798 here?

vdjeric: Pretty sure hard-coding a Y value is bad.. http://bit.ly/1FE0lOI

Uberi: Oops, should've tested with some smaller values as well - should be fixed as of ac7c5632d12bbb8af7c98d966738e13562a58f08.

PR #81, opened 2015-05-26
Add time range slider.

Uberi:
Now live at http://anthony-zhang.me/telemetry-dashboard


vdjeric:
@vitillo is this the kind of slider you wanted or did you want something like the old slider?

vitillo:
We are getting there :) When I move the slider the distribution is updated in real-time but it doens't feel smooth as in the old dashboard. Would it be possible to transition smoothly from one distribution to the other?

vitillo:
Sliding the distributon makes the window scroll up to the top. I have also seen the size of the x axis change while sliding, which shouldn't happen.

Uberi:
* I can't seem to make it scroll to the top when sliding - which browser/version does this occur for?
* Unfortunately, it would not be possible to transition between distributions unless we rolled our own graphing solution like before, with D3.
* The sliding issue seems to be fixed now - I can no longer reproduce it in the latest version.

vdjeric:
@vitillo merge it?

vitillo:
- If I scroll down as much as possible and then use the slider to change the distribution, the page scrolls slightly upwards.
- The x-axis can change while sliding. I can see why it makes sense to variate the limits of the y-axis but the x-axis should be fixed.

vitillo:
- Can we remove the imputed labels expressed in hours between the submission dates? I find it confusing. 
- When only a day is selected, the labels display numbers between 0 and 1000 in the upper graph. Can we remove those labels as well?
- Could we make the slider visually appealing? The label on the slider could be displayed on top of it and the slider could be extended to the right to align with the plots. Some UX engineer might provide guidelines on how a slider should look like.

Uberi:
* Fix scrolling upwards at the bottom of the page when the date range is adjusted.
* Fix the X axis changing when dragging.
* Remove the time labels in the evolution plot.
* Improve the appearance of the range slider.
* Remove the millisecond labels in the evolution plot.

vitillo:
The x-axis does still shift sometimes when dragging the time slider around. I think it's due to the size of the labels on the y-axis. The more space the labels occupy, the more the plot moves to the right. We should have a fixed spacing for the y axis labels.

Uberi:
* Fix the X axis shifting around when dragging the slider or changing the date.
    * Unfortunately, this required patching the library, so now we include a local copy.
* Support very large number value formatting using scientific notation.

vitillo:
This looks better, thanks. Something else is missing though. The histogram height should be equal (+/- fixed offset) to the bucket with the highest bar, i.e. the top most y label should be on par with the value of the bucket with the highest bar. This makes it easier to compare visually histograms while sliding over the timeline.

Uberi:
* Properly filter the dates in the summary range.
* Calculate the Y-axis scale manually to get rid of inconsistent spacing.
* Use a more sane date format in the date range selector.

vitillo:
@Uberi, did you update the version on http://anthony-zhang.me/telemetry-dashboard? I don't see any changes. I have also noticed that the plots and the slider don't appear in Chrome while the slider doesn't appear in Safari.

Uberi:
Should be updated now. Also added a workaround for a bootstrap bug (twbs/bootstrap#9382), so the range slider should properly show up now.

vitillo:
Great! Do you know why the dashboard doesn't display properly in other browsers?

Uberi:
Ah, that was Webkit treating `display: inline-table` as a block element when it should be an inline element. The commit above should make it work properly now (tested in Chrome 43 and IE 11).

vitillo:
Looks good, @vdjeric feel free to merge it.

PR #85, opened 2015-07-02
Add warning that recent versions seem to be missing submissions.

Uberi:
untitled


vitillo:
Could you link the bug as well in the warning (Bug 1179751)? Thanks.

PR #90, opened 2015-07-07
Feedback

Uberi:
Implement all the feedback currently at https://etherpad.mozilla.org/new-telemetry-dash-feedback.

Snapshot:

* Probe Search in Evolution Dashboard isn't very fuzzy - mconley
* Evolution dash:
    * plotting all versions of release channel 22 to 39 doesn't plot everything
    * release channel evo should have a prompt to switch x-axis to "by date" instead of "by buildID"
    * it should be possible to see the mean for boolean histograms
    * URL:  http://mzl.la/1KL4WC2
    * clicking on Feedback should open a new tab
    * going to the url above, clicking feedback, and hitting back results in "no data available"
    * reported by Markus Stange

### DIFF
--- a/dist.html
+++ b/dist.html
@@ -20,8 +20,8 @@
 <body>
     <div class="container-fluid">
         <header>
-            <form id="histogram-filters" class="navbar-form inline" style="float: right; margin: 0">
-                <a href="https://etherpad.mozilla.org/new-telemetry-dash-feedback" class="btn btn-success" title="Send feedback about Telemetry dashboards">Send feedback</a>
+            <form class="navbar-form inline" style="float: right; margin: 0">
+                <a href="https://etherpad.mozilla.org/new-telemetry-dash-feedback" class="btn btn-success" title="Send feedback about Telemetry dashboards" target="_blank">Send feedback</a>
                 <a href="./evo.html" class="btn btn-default" id="switch-views" title="Switch to the evolution view">Evolution View</a>
                 <a href="./" class="btn btn-default btn-primary" title="Landing page for Telemetry dashboards">Dashboards Home</a>
                 <a href="./tutorial.html#HistogramDashboard" class="btn btn-default" id="tutorial" title="Usage tutorial for Telemetry dashboards">Usage Tutorial</a>
--- a/dist.html
+++ b/dist.html
@@ -20,8 +20,8 @@
 <body>
     <div class="container-fluid">
         <header>
-            <form class="navbar-form inline" style="float: right; margin: 0">
-                <a href="https://etherpad.mozilla.org/new-telemetry-dash-feedback" class="btn btn-success" title="Send feedback about Telemetry dashboards" target="_blank">Send feedback</a>
+            <form class="navbar-form inline pull-right">
+                <a href="https://etherpad.mozilla.org/new-telemetry-dash-feedback" class="btn btn-success" title="Give feedback about Telemetry dashboards" target="_blank">Give feedback</a>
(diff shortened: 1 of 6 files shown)


dist.html
```diff
@@ -94,8 +94,8 @@ <h3 id="measure-description" style="font-size: 14px; margin: 0 3px"></h3>
                                     <label><input name="cumulative-toggle" type="radio" value="1"> Cumulative</label>
                                 </div>
                                 <div>
-                                    Date range variable: filter submissions by build ID range or submission date range:
-                                    <label><input name="build-time-toggle" type="radio" value="0"> Build Date</label>
+                                    Date range variable: filter submissions by build/release date range or submission date range:
```
vdjeric: why "release date"?

dist.html
```diff
@@ -94,8 +94,8 @@ <h3 id="measure-description" style="font-size: 14px; margin: 0 3px"></h3>
...
+                                    <label><input name="build-time-toggle" type="radio" value="0"> Build/Release Date</label>
```
vdjeric: ditto

evo.html
```diff
@@ -20,7 +20,7 @@
     <div class="container-fluid">
         <header>
             <form class="navbar-form" style="float: right; margin: 0">
-                <a href="https://etherpad.mozilla.org/new-telemetry-dash-feedback" class="btn btn-success" title="Send feedback about Telemetry dashboards">Send feedback</a>
+                <a href="https://etherpad.mozilla.org/new-telemetry-dash-feedback" class="btn btn-success" title="Send feedback about Telemetry dashboards" target="_blank">Send feedback</a>
```
vdjeric: Send feedback -> Give feedback

index.html
```diff
@@ -23,6 +23,9 @@
 <body>
     <div class="container-fluid">
         <header>
+            <form class="navbar-form inline" style="float: right; margin: 0">
```
vdjeric: don't hard-code style in the html

src/evo.js
```diff
@@ -53,20 +53,9 @@ Telemetry.init(function() {
...
+        $("#measure-description").text(description + " (" + measure + ")");
         $("#submissions-title").text(measure + " submissions");
-        
-        // Figure out which aggregates actually apply to this measure
-        var options;
-        if (measureEntry.kind == "linear" || measureEntry.kind == "exponential") {
-          options = [["median", "Median"], ["mean", "Mean"], ["5th-percentile", "5th Percentile"], ["25th-percentile", "25th Percentile"], ["75th-percentile", "75th Percentile"], ["95th-percentile", "95th Percentile"]]
-        } else {
-          options = [["submissions", "Submissions"]];
-        }
```
vdjeric: I think we still want this. Mean makes sense for flag, boolean histograms, the percentiles don't. Similarly it makes no sense to show Evo chart for enumerated histograms for example

dist.html
Uberi: For viewing data under the release channel, it might make more sense to say "release date" rather than "build date".

vdjeric: it really is a build date though. and release date & build date aren't
always the same. plus it's not data only collected from a specific release
day

vdjeric: maybe we can resolve this by putting "(switch to submission date)" or
"(switch to build ID)" next to the X-axis label on evo charts

Uberi:
Updated, now live at http://production.telemetry-dashboard.divshot.io/evo.html

vdjeric:
Merging, but there's another bug: when I open the dash in a private window (no cookies) and switch the histogram from GC_MS (exponential) to SHUTDOWN_OK (boolean), it says "Data missing" until I select Mean as the chosen measure. if there is only on emeasure available (Mean), it should choose it automatically

PR #91, opened 2015-07-08
Modified settings

Uberi:
* The advanced settings panel will now show "(modified)" when we change advanced settings, and open on load.
* It is no longer possible to select versions that are in different channels in evolution dashboard, like beta to nightly.
* The X-axis label now changes along with the evolution variable.
* Bugfixes for switching between aggregates of different kinds - switching to a flag histogram should automatically select the Mean aggregate, even when it was not previously selected.

Now live on http://production.telemetry-dashboard.divshot.io/

### DIFF
--- a/dist.html
+++ b/dist.html
@@ -83,7 +83,7 @@
             <div class="col-md-12">
                 <div class="panel panel-default">
                     <div class="panel-heading" role="tab">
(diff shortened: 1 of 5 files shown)


vdjeric:
Looks good!

PR #93, opened 2015-07-09
New UI fixes

Uberi:
* Fix bad versions showing up in the plots, such as beta/_7, even though they aren't visible from the selectors.
* Store the entire OS in the OS selector, which fixes the "OS versions randomly change when the measure or versions change" issue.
* Make the search bar in selectors stay at the top without wobbling, and focus it when we first open the dropdown. Improve search a bit.
* Charting and export fixes for the distribution dashboard, in particular for boolean and flag histograms.
* Various other tweaks and bugfixes.

Edit: latest version is now live at http://production.telemetry-dashboard.divshot.io/


src/dashboards.js
```diff
@@ -13,13 +13,16 @@ $(document).ready(function() {
...
       disableIfEmpty: true,
       nSelectedText: $this.data("n-selected") !== undefined ? $this.data("n-selected") : "selected",
+      onDropdownShow: function(event) { // Focus the search box whenever the dropdown is opened
+        setTimeout(function() { $(event.currentTarget).find(".filter input").focus(); }, 0);
+      },
     };
     if ($this.attr("id") === "measure") {
       options.filterBehavior = "custom";
       options.filterCallback = function(element, query) {
         var value = $(element).find("label").text().toLowerCase();
```
vdjeric: value is not a good variable name here, is this the search string or one of the options in the dropdown list?

src/dashboards.js
```diff
@@ -176,8 +175,7 @@ function getHumanReadableOptions(filterName, options, os) {
   } else if (filterName === "measure") {
     return options.map(function(option) { return [option, option] });
   } else if (filterName === "channelVersion") {
-    var pattern = /^\w+\/\d+$/;
-    return options.filter(function(option) { return pattern.test(option); }).map(function(option) { return [option, option.replace("/", " ")] });
+    return options.map(function(option) { return [option, option.replace("/", " ")] });
```
vdjeric: arrow function syntax is cleaner :)

src/dashboards.js
```diff
@@ -323,4 +323,41 @@ var indicate = (function() {
...
-})();
\ No newline at end of file
+})();
+
+function compressOSs() {
+  var selected = $("#filter-os").val() || [];
+  var options = $("#filter-os option").map(function(i, element) { return $(element).attr("value"); }).toArray();
+  var optionCounts = {}, selectedByOS = {};
+  options.forEach(function(option) {
+    var os = option.split(",")[0]; optionCounts[os] = optionCounts[os] + 1 || 1;
```
vdjeric: don't squeeze in multiple statements on a single line

src/dashboards.js
```diff
@@ -323,4 +323,41 @@ var indicate = (function() {
       $(".busy-indicator").hide();
     }
   }
-})();
\ No newline at end of file
+})();
+
+function compressOSs() {
```
vdjeric: hmm, so there might be a simpler way to do this.. in the OS box, have the top 3 options be, "Windows (all)", "Linux (all)", "OS X (all)".. and then some kind of divider followed by the actual os/version combos as additional options. And if a person manual selects all windows versions but doesn't check the "windows" box, then don't bother trying to "compress" it into os=WINNT.. it will be super-rare for someone to hurt themselves that way :)
Also, selecting a specific e.g. Windows version should clear the "All Windows" option (but not the All Linux or All OS X options"

src/dashboards.js
Uberi: That would be ideal, but it doesn't work in Chrome - it's an ES6 feature.

vdjeric: nuts

src/dashboards.js
vdjeric: i see you decided to keep compressOSs -- I guess that works too. But can you add some kind of indicator that the "Windows"/"Linux"/"Darwin" *labels* are clickable.. like maybe "Windows (click for all)"?

Also if the user has selected by OS (e.g. WINNT), it shouldn't say "7 OSs", it should say "Windows".
Alternately it should say "Windows, OS X" or "Windows, OS X, Linux" if two or three entire OS familes are selected

vdjeric:
- This patch doesn't merge cleanly (see github warning)
- Bug: Side-box doesn't match advanced settings box: http://bit.ly/1Mq5Otf
- Bug: Adding "arm" to http://bit.ly/1Mq6iQ3 somehow changes distribution, but arm alone has zero submissions??
- Filename chosen by "Export CSV" is terrible: "function Histogram_measure()..."
- Evo dash: Mouse cursor turns into a caret over "click to use Build Id" instead of a click-hand? Change the "click to use build ID" text portion into a blue hyperlink

vdjeric:
Bug: this is broken http://bit.ly/1Mq7bIu

Uberi:
* Merged from master.
* Disregard invalid start and end dates.
* Properly detect when all the options of a selector are selected after library upgrade (causes the selector to think it's all selected when all but 1 are).
* Fix the advanced dashboard spinning forever if the cookie is invalid and there is no URL state, which could happen if visiting an invalid permalink.
* Fix weird export file names.
* Make the X-axis of the evolution dash clickable.

Deployed to http://production.telemetry-dashboard.divshot.io/.

vdjeric:
*Bugs*:

- Date-picker and legend don't agree on the number of days: http://bit.ly/1I8pw8I
- Can't create link to GC_MAX_PAUSE_MX for Release 30-33 with 64-bit on Windows only

*Further improvements:*

General:
- Need a table-view for dist
- Implement rest of suggestions from https://etherpad.mozilla.org/new-telemetry-dash-feedback

Channels:
- The search box for Firefox channel is still too small on opening
- The top items in the box should be: latest nightly, latest aurora, latest beta, latest release and they should point to the current version
- we shouldn't put the channels with very few submissions (e.g. nightly 0, nightly 1, nightly 2) at the bottom of the list.. that way we can notice that something is wrong, but people won't waste time scrolling the list past them
- sort the list by version number, so: latest nightly (42), latest aurora (41), latest beta (40), latest release (39), then a separator, then nightlies, auroras, betas, releases
- Channel selector should should say "Windows" vs "10 OSs"

src/dashboards.js
```diff
@@ -323,4 +349,42 @@ var indicate = (function() {
...
+  var selected = $("#filter-os").val() || [];
+  var options = $("#filter-os option").map(function(i, element) { return $(element).attr("value"); }).toArray();
```
vdjeric: Break the inline function into multiple lines

PR #94, opened 2015-07-10
Add histogram simulator

Uberi:
Currently live at http://development.telemetry-dashboard.divshot.io/histogram-simulator/


vdjeric:
r-, needs more testing:
- I see stuff like "8.64 samples" (is it 8.64%)
- right-most bucket (overflow bucket) is clipped
- up-arrow in the text-boxes is clipped
- you didn't add the terminology from HIstogram.json
- X-axis label is to the right of the X-axis instead of to the left
- etc etc

vdjeric:
- add the "kind" label
- Fix bucket ranges (no decimals)
- Don't generate negative values for the distributions
- Why is'nt the label for bucket sample-count & bucket-range not over the bar, like in the real dash?

vdjeric:
- When I load the page http://development.telemetry-dashboard.divshot.io/histogram-simulator/ I see low=1,high=500,n_buckets=20 for a flag histogram? Flag histograms don't have n_buckets or high or low. Only show the valid histogram parameters
- It's also unclear which distribution you used to generate the initial histogram on page-load (uniform?).
- Can you make the distribution types radio-buttons instead of buttons
- I don't think the "to plot this data" box makes sense.. no one wants to manually inspect 10,000 values or copy-paste them into a spreadsheet. For now, just sort the values in the box
- Changing histogram values doesn't automatically regenerate the histogram
- low=1, high=10, n_buckets=20 generates a weird looking distribution. maybe warn when the parameters don't make sense?
- linear histogram with normal distribution + low=1,high=10k,n_buckets=100 looks super odd (especially the last bar).. are you generating the normal distribution properly?
- change TEST_HISTOGRAM x-axis label to "Buckets"
- red labels overlap the text-boxes

vdjeric:
A shareable link would be good too.. i.e. update the location bar

Uberi:
* Only show the valid histogram parameters.
* Make the distribution types radio-buttons instead of buttons.
* Sort the values.
* Changing histogram values automatically regenerates the histogram.
* Warn when the parameters don't make sense.
* Change TEST_HISTOGRAM x-axis label to "Buckets".
* Red labels should not overlap the text-boxes.

Also deployed to the link above, and added support for shareable links in the URL bar.

vdjeric:
The landing page is still a flag histogram with 20 buckets, a high of 500, etc?

vdjeric:
"Use this histogram.." no longer makes sense

Uberi:
The inputs are disabled when flag or boolean histograms are selected, but they retain their value in case the user switches back to the linear or exponential histogram types (the disabled fields are not used in the bucket generation).

vdjeric:
Can you warn users if they try to declare a linear histogram with buckets <0, 1, 2, .., N >?
It's in the Telemetry docs:  "NOTE: If you need a linear histogram with buckets < 0, 1, 2 ... N >, then you should declare an enumerated histogram. This restriction was added to prevent developers from making a common off-by-one mistake when specifying the number of buckets in a linear histogram."

vdjeric:
Changing a field (such as n_buckets) doesn't immediately update the graph. Graph only updates when I take focus away from the text box. This can be misleading because text-box will show the new value while graph is still showing the old value

vdjeric:
This link has n_buckets=12 but the graph only shows 11 buckets:
http://development.telemetry-dashboard.divshot.io/histogram-simulator/#low=0&high=100&n_buckets=12&kind=linear&generate=uniform

We should show the overflow bucket. We should also indicate that the first bucket is the underflow bucket and the last bucket is the overflow bucket in the tooltips.

vdjeric:
Thanks!
More feedback from Ehsan's UX testing:
- In top-left corner of the histogram, or in a label above it(?), show the bucket "width"
- Explain the meaning of the overflow & underflow buckets.. maybe in a label above "Histogram Properties"? 
- Put the histogram "kind" as the first item
- Is upper_bound inclusive? when a developer specifies high=100, does that mean a sample value of 100 will go into the overflow bucket? that's what the simulator suggests but is that what really happens in the code? Can you also check if we always have a separate bucket for underflow & overflow?
You can test this out in Firefox -- load about:telemetry, open DevTools console on the page and accumulate samples: e.g. Services.telemetry.getHistogramById("GC_MS").add(123)
Then refresh about:telemetry to test what the resulting histogram looks like

vdjeric:
The bucket boundaries need to be more visible, can we put a permanent label on all the histogram bars?

Uberi:
* Bucket width and bucket descriptions.
* Add description for underflow and overflow buckets.
* Kind should be the first item in the histogram settings.
* The upper bound is specified to be exclusive - `high` values of 1000 mean that values of xactly 1000 go into the overflow bucket (confirmed on various histograms in `about:telemetry`).
* Add permanent labels to the bucket boundaries.

histogram-simulator/index.html
```diff
@@ -0,0 +1,90 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    
+    <title>Histogram Simulator</title>
+    
+    <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css">
+    <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
+    <link rel="stylesheet" type="text/css" href="/style/metricsgraphics.css" />
+    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/codemirror/5.4.0/codemirror.css">
+    <style>
+.CodeMirror {
```
vdjeric: why don't you put this in a CSS file?

vdjeric: indentation is also off?

vdjeric:
- Why can't I have a histogram with 10 buckets 1 t0 10? This is as close as I could get: http://development.telemetry-dashboard.divshot.io/histogram-simulator/#low=1&high=11&n_buckets=10&kind=linear&generate=custom
Is this how the Telemetry code behaves?
- It's impossible to type into the custom-data box, I can only paste (try it)

Uberi:
* This was not possible due to that being equivalent to an enumerated histogram - there is a check for if the user makes a linear histogram having buckets `1, ..., N`. The above changes now make that possible, using `low=1,high=10,n_buckets=11`.
* This is now fixed, and all the above changes have been deployed to the divshot link above.

Uberi:
* `histogram_tools.py` also allows creation of a linear histogram with buckets `1,2,3,...N`.
* In fact, it is also possible to create one with more buckets than distinct integer values in the range: `0,1,2,3,4,5,6,6,7,8,9,10` for `linear_buckets(1, 10, 12)`. This probably breaks a lot of things everywhere else though.
* The simulator uses a line-by-line translation of the bucket calculations in `histogram_tools.py`, and the results should be the same.
* It's probably best to disallow having more buckets than distinct integer values in the range, but allow having the same number of buckets as possible integers (like it is in the latest set of changes).

vdjeric:
The simulator page should put up a red-text warning (or something similar) when the user has chosen values that don't make sense, but it should only *prevent* configurations that are forbidden by the Histograms.json parsing scripts

vdjeric:
can you add a mention that the "low" paramter is option and that the default value is 1

PR #97, opened 2015-07-21
Bugfixes 2

Uberi:
Implements fixes and features from from #93:

* Simpler timezone handling - just convert it into a date string (this also fixes the legend and the date range picker getting out of sync).
* Make the search box bigger by removing the unnecessary icons.
* Sort the version selector better and add separators to make the list easier to read.
* OS selector shows the OSs themselves when they are all selected, like "Windows" when all the Windows versions are selected.
* Update the finished feedback items.

Still to be done:

* Table view for dist dash.
* Add option to select an entire channel at once in the evolution dash.

Now live at http://production.telemetry-dashboard.divshot.io/.

PR #102, opened 2015-08-06
Various fixes from #97

Uberi:
* Don't hardcode max version - default to latest nightly.
* When changing sort order of keys in a keyed histogram, reset the four distribution graphs to the top 4 again.
* Filter out version numbers that seem to be too high in the version selector in the evolution dash.
* Apply settings such as cumulative-ness to the export files.
* Pretty-print the JSON in export files.
* More contrast for the lines in graphs.
* Limit select-all select single option behaviour to product and OS selectors (elsewhere, they are confusing to use).
* Use shorter names for OS selector tooltip.
* Improve the product list display.

These changes have been applied to both the v2 and v4 dashboards. The changes are live at http://production.telemetry-dashboard.divshot.io/.

### DIFF
--- a/dist.html
+++ b/dist.html
@@ -89,6 +89,11 @@
                         <div class="panel-body">
                             <form class="form-horizontal">
                                 <div>
+                                    <b>Table mode:</b> show histograms as tables of values rather than visual charts:
+                                    <label><input name="table-toggle" type="radio" value="0"> Show histograms</label>
+                                    <label><input name="table-toggle" type="radio" value="1"> Show tables</label>
+                                </div>
+                                <div>
                                     Cumulative mode: show each bucket as the sum of itself and all buckets on its left:
                                     <label><input name="cumulative-toggle" type="radio" value="0"> Non-cumulative</label>
                                     <label><input name="cumulative-toggle" type="radio" value="1"> Cumulative</label>
--- a/new-pipeline/src/dist.js
+++ b/new-pipeline/src/dist.js
@@ -500,6 +500,8 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
 }
 
 function displaySingleHistogramSet(axes, useTable, histograms, title, cumulative, trimLeft, trimRight, maxPercentage) {
+  $(axes).empty(); // Remove tables if they are present
+
   // No histograms available
   if (histograms.length === 0) {
     MG.data_graphic({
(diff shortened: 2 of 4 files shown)


vdjeric:
**Bug fix:**
- "TypeError: pageState.min_channel_version is null" in dashboards.js:177 on Histogram V4: http://production.telemetry-dashboard.divshot.io/new-pipeline/dist.html#!compare=osVersion&cumulative=0&end_date=2015-08-04&max_channel_version=beta%252F40&measure=BROWSER_IS_USER_DEFAULT&min_channel_version=null&os=Windows_NT%252C5.2!Windows_NT%252C6.0!Windows_NT%252C6.3!Windows_NT%252C5.1!Windows_NT%252C6.1!Windows_NT%252C10.0!Windows_NT%252C6.2&sanitize=1&sort_keys=submissions&start_date=2015-07-02&table=0&trim=1&use_submission_date=0
- What is Linux 13-22? Are those Android versions? We should have an "All Android" selector as well as the descriptive names, e.g. "Android 5.0 Lollipop"
- What's the target OS for getting B2G data to show up? I can't get any B2G data
- V4 histogram dashes

**Enhancement:**

- Export JSON/CSV should export the % for each bucket as well
- Histogram date selector: "Last 7 days" = Last 7 days of data?
- Can we fix the DevTools console warning: "WARNING: telemetry.js is loaded in development mode"
- Why does http://bit.ly/1TgVgnM repeat the name of the histogram in the caption? "Longest GC slice in a GC (ms) (GC_MAX_PAUSE_MS)"

**Other:**
- http://telemetry.mozilla.org/old/ is broken, Fix it if it's easy or get rid of it?
- Rename "Advanced dashboard" (http://telemetry.mozilla.org/advanced/) to "Old dashboard" in t.m.o. Get rid of the Nightly 39/40 data disclaimer at the top

vdjeric:
Go to V4 evo dash, select the boolean histogram "BROWSER_IS_USER_DEFAULT", the default measure is "Median"??

vdjeric:
V4 dash -> BROWSER_IS_USER_DEFAULT -> change "from" date to release 39. End date ends up release 4??

Run the V4 dash through your test plan

Uberi:
* Divshot deploy failed somehow... should be the actual latest version now (caused the TypeError earlier).
* "Linux" appears to include both desktop Linux and Android, and there's no way to distinguish between them besides possibly guessing versions.
* B2G submissions show up when sanitize is turned off - we don't seem to have any data for this.
* Added percentages to exports.
* Console warning fixed.
* Repeating the histogram name was added because, back when it wasn't, people would get confused as to what the description actually described - the entire sentence, or just the measure itself. Using tooltips would also be a solution, but then most people wouldn't know it was there unless they accidentally mouse over it.
* Classic dashboard removed, advanced dashboard renamed to "Old Dashboard".
* Bugfixes for test plan results.
* Automatically switch to mean for flag and boolean histograms.
* Fix invalid version ordering in channel switching.

Issue #115, opened 2015-08-15
Layout bug in anthony-zhang.me v4 dash. Is t.m.o dash affected?


vdjeric:
Example: http://bit.ly/1W2Wv96

vdjeric:
Confirmed: http://mzl.la/1fcyISs
Please fix :)

Uberi:
Ref #119

PR #119, opened 2015-08-17
Fix layout issue with keyed histograms not rendering properly.

Uberi:
See #115 for details.

The third keyed histogram plot aligns to the right instead of the left, and messes up the layout.


vdjeric:
LGTM, i'll wait for merge conflicts

Issue #117, opened 2015-08-15
V4 histogram dash shouldn't have "key sorting" option for unkeyed histograms


vdjeric:

Uberi:
Ref: #121

PR #121, opened 2015-08-17
Only show sort keys option for keyed histograms.

Uberi:
Ref: #117

### DIFF
--- a/new-pipeline/src/dist.js
+++ b/new-pipeline/src/dist.js
@@ -120,9 +120,9 @@ $(function() { Telemetry.init(function() {
             return getAggregate(entry2.histograms) - getAggregate(entry1.histograms);
           });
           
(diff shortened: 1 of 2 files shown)


new-pipeline/src/dist.js
```diff
@@ -120,9 +120,9 @@ $(function() { Telemetry.init(function() {
             return getAggregate(entry2.histograms) - getAggregate(entry1.histograms);
           });
           
+          var keys = gCurrentHistogramsList.map(function(entry) { return entry.title; });
```
vdjeric: arrow functions are coming to Chorme 45 (currently supported in Chrome Beta, launching in a week) and Edge supports them already: https://www.chromestatus.com/feature/5047308127305728

I'm tempted to ask you to switch to arrow functions in advance.. it reads a lot cleaner

PR #124, opened 2015-08-18
Change BASE_URL from http to https

bifurcation:
Otherwise, this breaks on HTTPS sites.


vdjeric:
@Uberi can you check if we use HTTP anywhere else?

Uberi:
Ref: #125

PR #132, opened 2015-08-18
Support export within comparison view.

Uberi:
Allow export of comparison values.

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -85,27 +85,29 @@
                             </div>
                         </div>
                     </div>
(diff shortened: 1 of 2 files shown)


vdjeric:
Where can i see this?

Uberi:
Now live at http://executor-lionel-45300.bitballoon.com/new-pipeline/dist

(Note that this is hosted on BitBalloon, so the "Evolution View"/"Histogram View" buttons on the top bar will not work)

vdjeric:
See feedback in #136 and #137

PR #134, opened 2015-08-18
Better bucket labels.

Uberi:
Make pie chart bucket labels better.

They default to True, False, and Other now.


Uberi:
I couldn't find any, but I do remember seeing them in the past on the v2 pipeline for older versions.

PR #157, opened 2015-08-20
Invert boolean bucket pie chart labels. Fixes #156

### LINKED ISSUE #156: Boolean pie chart labels are inverted
For PWMGR_SAVING_ENABLED on Beta 40:
Start	End Count
0	1	1.55M (4.21%)
1	2	35.24M (95.79%)

Where 0 is false as (one would normally expect) yet the pie graph tells the opposite story saying that 95.79% have password saving disabled.

mnoorenberghe:
0 = False
1 = True


vdjeric:
Thanks for the pull :+1:

Issue #176, opened 2015-08-27
Probe selector loses its filter upon re-opening


mnoorenberghe:
When I re-open the probe dropdown after having selected a new probe, the popup initially opens with the list pre-filtered or at least scrolled but then switches to a unfiltered view from the top. Ideally the filter the should be remembered since one normally wants to look at related probes e.g. all `PWMGR_` ones

I think this may be a regression since I don't think this happened a month ago.

![telemetry_dash_filter_revert](https://cloud.githubusercontent.com/assets/601139/9510794/ce35ac84-4c24-11e5-978a-829cf50a866c.gif)

mnoorenberghe:
This is the newest evo dashboard btw.

vdjeric:
I actually asked for this in https://github.com/mozilla/telemetry-dashboard/issues/167
:)
Maybe we can compromise on having the box blank but the list cued to the right position?

mnoorenberghe:
That would work for the most common case where the same prefix is used so is better than the current behaviour. What was your use case for #167?

mnoorenberghe:
Another idea would be to revert #167 but have the filter text selected and focused so it's simple to type over it.

Uberi:
IIRC the original reason for #167 was the need to clear the filter box every time you wanted to change filters, which is an extra click/keystroke.

vdjeric:
Merged & deployed #177

PR #177, opened 2015-08-27
Make selectors scroll the last selected items into view.

Uberi:
Ref: #176.

Now live at http://planner-frog-12114.bitballoon.com/.

### DIFF
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -21,8 +21,7 @@ $(document)
               var selectedItem = container.find(".active");
 
               // Clear the search box and focus it
-              container.find(".filter input").val("").trigger("input");
-              container.focus();
+              container.find(".filter input").val("").trigger("input").focus();
 
               // Scroll the previously selected item into view
               setTimeout(function() {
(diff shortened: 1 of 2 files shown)


vdjeric:
There's a bug. STR:

1. Enter "pause" in the histogram name box
2. Select GC_MAX_PAUSE_MS histogram
3. Open histogram name box

Expected: list should be scrolled to GC_MAX_PAUSE_MS
Actual: the beginning of the list is in view

Uberi:
New version is now up at http://money-lender-mona-33020.bitballoon.com/.

vdjeric:
Can you do anything about the "blink"?

STR:
1. Type "uptime" in the histogram box. Observe how the list shrinks down to only one matching histogram
2. Select SIMPLE_MEASURES_UPTIME histogram
3. Open histogram name box again

Expected: histogram name box opens, cued to the currently selected histogram's name in the full list

Actual: histogram name box opens to its previous view first (only 1 histogram), then almost immediately resets to the correct view (full list cued to SIMPLE_MEASURES_UPTIME)

Uberi:
The blink is caused by the way the selector library works, so it would require library patching. Could be done, but would be pretty messy.

vdjeric:
Well I guess this blink isn't new behaviour, and the new functionality **is** better. Merged.

vdjeric:
Deployed too

PR #180, opened 2015-09-15
hacky fix to return a bare evolution object if the histogram is not keyed

canuckistani:
Most data in telemetry currently is unkeyed, and it feels awkward to have to refer to the evolutions returned via getEvolution with this `evolutionMap[""].map(...)` - so I added what seems to be reasonable detection of this case in getEvolution and return a bare Evolution object instead

This isn't a great solution because the argument to the callback can now be one of two different things. An alternative might be to have a convention such as 'evolutionMap.evolution' for unkeyed histograms?

### DIFF
--- a/v2/telemetry.js
+++ b/v2/telemetry.js
@@ -483,7 +483,7 @@
             metric);
         };
         // Detect un-keyed histograms and return the unwrapped evolution
-        if (Object.keys(entriesMap).length === 1 && entriesMap[0] === void 0) {
+        if (Object.keys(entriesMap).length === 1 && typeof entriesMap[0] === 'undefined') {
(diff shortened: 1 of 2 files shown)


v2/telemetry.js
```diff
@@ -482,6 +482,10 @@
             entriesMap[label], histograms.kind, histograms.description,
             metric);
         };
+        // Detect un-keyed histograms and return the unwrapped evolution
+        if (Object.keys(entriesMap).length === 1 && entriesMap[0] === void 0) {
```
vdjeric: can you change "void 0" to "undefined", it's kinda rare :)

vdjeric:
@canuckistani what did we end up deciding about this pull request?

PR #188, opened 2015-11-02
bug 1219790 - Accept a &bare=true parameter for no-frills viewing. r?vladan

chutten:
No-frills means no titles, no query strings, no comments and, in the case of evolutions, no submission data.

To use: just set your src to be the url you want plus &bare=true.

It doesn't load very prettily because it doesn't read the options until it has mostly finished rendering.

This is just a quick'n'dirty way to get this capability. To do this more sensibly will require some refactoring of the dashboard code (bug 1220185)

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -19,7 +19,7 @@
 </head>
 <body>
     <div class="container-fluid">
-        <header>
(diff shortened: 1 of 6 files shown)


chutten:
Went a different way, this is not needed.

PR #191, opened 2015-11-10
Ignore subsequent calls to v2/telemetry.js' Telemetry.init().

chutten:
If you have a long-lived script, you will have to reload the page so that
Telemetry.init can be called again when another date's data is available.

Without this, calling Telemetry.init over and over again could cause callbacks
to be called when the internal version->date structures are in an inconsistent
state.

### DIFF
--- a/v2/telemetry.js
+++ b/v2/telemetry.js
@@ -368,8 +368,24 @@
     xhr.send();
   }
 
+  var _init;
+  var _initCompletedCallbacks;
   Telemetry.init = function Telemetry_init(callback) {
     assert(typeof callback === "function", "`callback` must be a function");
+
+    if (_init) {
+      callback();
+      return;
+    }
+    if (_initCompletedCallbacks) {
+      _initCompletedCallbacks.push(callback);
+      return;
+    }
+    _initCompletedCallbacks = [
+      function () { _init = true; },
+      callback,
+    ];
+
     Telemetry.getJSON(Telemetry.BASE_URL +
       "aggregates_by/build_id/channels/",
       function (channels) {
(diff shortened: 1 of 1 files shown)


vdjeric:
@mreid-moz any thoughts?

v2/telemetry.js
```diff
@@ -368,7 +368,13 @@
     xhr.send();
   }
 
+  var _init;
   Telemetry.init = function Telemetry_init(callback) {
+    if (_init) {
+      callback();
```
mreid-moz: This should be moved below the assert that checks that `callback` is a function

mreid-moz:
Hmm... if you called the callback twice, wouldn't you still get inconsistent state the second time (since the initialization wouldn't yet be complete)?

It might be better to simply return without calling the callback in the case where initialization has already happened.

chutten:
You have to call the callback every time, otherwise the caller won't know to proceed.

But you're right. Two quick calls, and the second might return before the first. Ideally we'd wrap these calls as promises and have any callers then() against them, but I'm trying _really_ hard to make as few intrusions as possible.

mreid-moz:
For a minimally invasive change, I'd suggest this:
- keep another small piece of state for "initialization in progress"
- if `init` is called again while ^^ is true, throw an informative exception
- otherwise call the callback as expected.

Then you have a small chance at a debuggable error, otherwise things are consistent.

That seems slightly preferable to a small chance at inconsistent state, with no error at all.

chutten:
For slightly more trouble I can keep an array of callbacks and call them all in order once the first init's done. Bonus: no exception. Let me play with the code...

mreid-moz:
Yeah, I was going to suggest that too, but didn't want to send you down a rabbit hole of race conditions and complexity for something that will (hopefully) rarely/never be needed :)

vdjeric:
@chutten still want me to merge this?

chutten:
@vdjeric yes please.

PR #196, opened 2015-11-18
dashboard-generator: To-order, on-demand, hot and cold running dashboards

chutten:
The blogpost isn't live yet, but it has been written.

### DIFF
--- a/dashboard-generator/generator.js
+++ b/dashboard-generator/generator.js
@@ -23,13 +23,18 @@ window.addEventListener('load', function () {
   var _versions; // {channel: [versions (sorted)], ...}
   var _dash = []; // ordered list of plots for your dash
 
+  const VERSIONS_OFF_NIGHTLY = {
+    'nightly': 0,
+    'aurora': 1,
+    'beta': 2,
+    'release': 3,
+  };
+
   Telemetry.init(() => {
 
     // versions are static across the session, so stash 'em
     _versions = {};
     Telemetry.getVersions()
-      .sort()
-      .reverse()
       .map(versionString => {
         var [channel, version] = versionString.split('/');
         if (!_versions[channel]) {
@@ -38,6 +43,17 @@ window.addEventListener('load', function () {
           _versions[channel].push(version);
         }
       });
+    // aggregates.t.m.o doesn't sanitize its versions well
+    _versions['nightly'].sort((a, b) => b - a);
+    for (var channel in _versions) {
+      if (channel == 'nightly') {
+        continue;
+      }
+      var maxVer = _versions['nightly'][0] - VERSIONS_OFF_NIGHTLY[channel];
+      _versions[channel] = _versions[channel]
+        .filter(ver => ver <= maxVer)
+        .sort((a, b) => b - a);
+    }
 
     $('#channel').addEventListener('change', updateVersions);
     $('#version').addEventListener('change', updateMetricsAndComparesAndAppsAndOS);
--- a/dashboard-generator/generator.js
+++ b/dashboard-generator/generator.js
@@ -1,0 +1,361 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// utility functions
+var $ = (selector) => document.querySelector(selector);
+
+var createOption = (parent, value, text = value, selected = false) => {
+  var option = document.createElement('option');
+  option.value = value;
+  option.textContent = text;
+  option.selected = selected;
+  parent.appendChild(option);
+};
+
+var removeAllChildren = (el) => {
+  while (el.firstChild) {
+    el.removeChild(el.firstChild);
+  }
+};
+
+window.addEventListener('load', function () {
+  var _versions; // {channel: [versions (sorted)], ...}
+  var _dash = []; // ordered list of plots for your dash
+
+  Telemetry.init(() => {
+
+    // versions are static across the session, so stash 'em
+    _versions = {};
+    Telemetry.getVersions()
+      .sort()
+      .reverse()
+      .map(versionString => {
+        var [channel, version] = versionString.split('/');
+        if (!_versions[channel]) {
+          _versions[channel] = [version];
+        } else {
+          _versions[channel].push(version);
+        }
+      });
+
+    $('#channel').addEventListener('change', updateVersions);
+    $('#version').addEventListener('change', updateMetricsAndComparesAndAppsAndOS);
+    updateChannels();
+    updateVersions();
+    updateMetricsAndComparesAndAppsAndOS();
+    updateE10s();
+
+    $('#compare').addEventListener('change', () =>
+      $('#sensible-compare').disabled = !$('#compare').selectedOptions[0].value);
+
+    $('#add').addEventListener('click', addPlotToDash);
+
+    loadPlotsFromLocation();
+  });
+
+  function addPlotToDash() {
+    var plotParams = {
+      channel: $('#channel').selectedOptions[0].value,
+      version: $('#version').selectedOptions[0].value,
+      metric: $('#metric').value,
+      useSubmissionDate: $('#use-submission-date').checked,
+      sanitize: $('#sanitize').checked,
+      trim: $('#trim').checked,
+      compare: $('#compare').selectedOptions[0].value,
+      sensibleCompare: $('#sensible-compare').checked,
+      evoVersions: $('#evo-radio').checked ? $('#evo-versions').value : 0,
+      filters: undefined,
+    };
+
+    // now to add the filters
+    if ($('#show-filters').checked) {
+      var filters = {};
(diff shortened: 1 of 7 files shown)


vdjeric:
I think there's a bigger issue here.. neither the old EVENTLOOP_UI_LAG_EXP_MS nor its replacement EVENTLOOP_UI_ACTIVITY_EXP_MS histogram are showing up in the t.m.o dashboards for Aurora/Beta/Release, e.g. http://mzl.la/1kIbTtt
This might be an aggregator bug. Can you talk to @mreid-moz

chutten:
Will do.

vdjeric:
A few other comments:
* the varying-os functionality compares Windows_NT, Windows_95, Linux, Darwin. There should only be one Windows family
* the varying-os distribution graph only draws 1 line
* Filter info is not reflected anywhere in the graphs
* the first build date on an evo X-axis includes the year, and this overlaps with the month

tested with:
/* The generator that created this dash can be found at:
 * http://chutten.github.io/dashboard-generator/?channel=nightly;nightly;nightly;nightly;nightly;nightly;nightly;nightly;nightly;nightly&version=;;;;;;;;;&metric=SIMPLE_MEASURES_FIRSTPAINT;SIMPLE_MEASURES_FIRSTPAINT;SIMPLE_MEASURES_FIRSTPAINT;SIMPLE_MEASURES_FIRSTPAINT;SIMPLE_MEASURES_FIRSTPAINT;SIMPLE_MEASURES_FIRSTPAINT;SIMPLE_MEASURES_FIRSTPAINT;SIMPLE_MEASURES_FIRSTPAINT;SIMPLE_MEASURES_FIRSTPAINT;SIMPLE_MEASURES_FIRSTPAINT&useSubmissionDate=false;false;false;false;false;false;false;false;false;true&sanitize=true;true;true;true;true;true;true;true;true;true&trim=true;true;true;true;true;true;true;true;true;true&compare=e10sEnabled;os;os;os;os;os;os;os;os;os&sensibleCompare=true;true;true;true;true;true;true;true;true;true&evoVersions=0;0;3;3;3;3;0;0;0;3&filters={"application":"Firefox","os":"Windows_NT"};{"application":"Firefox"};{"application":"Firefox","os":"Windows_NT"};{"application":"Firefox","os":"Windows_NT","e10sEnabled":true,"child":false};{"application":"Firefox","os":"Windows_NT","e10sEnabled":true,"child":true};{"application":"Firefox","os":"Windows_NT","e10sEnabled":false,"child":false};{"application":"Firefox","os":"Windows_NT","e10sEnabled":false,"child":false};{"application":"Firefox","os":"Windows_NT","e10sEnabled":true,"child":false};{"application":"Firefox","os":"Windows_NT","e10sEnabled":true,"child":true};{"application":"Firefox","os":"Windows_NT"}
 */

chutten:
>the varying-os functionality compares Windows_NT, Windows_95, Linux, Darwin. There should only be one Windows family

Sensibly comparing OS doesn't actually just limit to three. It draws a line for each `OS` in the `OS,VER` strings returned by `getFilterOptions` for the selected channel and version (those that aren't sanitized out because they have too few submissions). This is for future extensibility. 

A lot of the problem plots have both an `os` setting on `filter` and `compare`. In t.m.o, `filter` wins out and limits the values available to `compare`, which might clear up some of the problems. 

Though, running some of the sample parameters through t.m.o shows that it's following a similar tactic wrt Windows versions: http://mzl.la/1QQrdjJ 

Regardless, I'll filter the compare values. If `compare` is set to a value that is present in `filters`, I'll suppress the `compare`.

>the varying-os distribution graph only draws 1 line

This is odd. It looks like it's drawing the sum, somehow. I'll take a look before I do the filter/compare work.

>Filter info is not reflected anywhere in the graphs

The titles might get a bit longwinded. "SIMPLE_MEASURES_FIRSTPAINT - with varying os - filtered to os="Windows_NT", e10sEnabled=true, child=true"
Maybe I could render the extra bits as subtitles...

>the first build date on an evo X-axis includes the year, and this overlaps with the month

Odd, I don't see this. Mind you, my evo plots only have three x-axis ticks, so it might be viewport-based. I'll see if I can reproduce on my own. If not, I may pester you for a screenshot.

chutten:
On the EVENTLOOP* front, discussion and investigation uncovered: Bug 1226337 - EVENTLOOP_UI_{LAG,ACTIVITY}_EXP_MS needs to become an "exceptional case" in histogram.py

chutten:
The single-line hist was due to the `filter` overriding the `compare` so it was displaying the Windows_NT value over and over. So that's fixed with the "Don't allow compares by values that have been filtered" commit.

There were some ordering problems uncovered with errors while testing.

Non-measure-name information is now communicated via a subtitle.

I'm using the https codepen POST url now to avoid a "over insecure connection" dialog

Here's the new demo page: https://people.mozilla.org/~chutten/telemetry-dashboard/dashboard-generator/index.html

vdjeric:
* Use medians instead of means 
* the lack of SIMPLE_MEASURES_FIRSTPAINT in child processes is an aggregator bug, I filed a bug https://bugzilla.mozilla.org/show_bug.cgi?id=1228126
* I think people will want a provision for custom titles, but let's cross that bridge when we get to it

vdjeric:
* Also, this Beta evo graph shows Beta versions which are not released yet + only shows 3 versions instead of 4: https://people.mozilla.org/%7Echutten/telemetry-dashboard/dashboard-generator/index.html?channel=beta&version=&metric=GC_MAX_PAUSE_MS&useSubmissionDate=false&sanitize=true&trim=true&compare=&sensibleCompare=true&evoVersions=4&filters=
* Additionally that generator link above never finishes loading

chutten:
>Also, this Beta evo graph shows Beta versions which are not released yet + only shows 3 versions instead of 4

Odd,
```js
Telemetry.getVersions().filter(version => version.startsWith('beta'))
Array [ "beta/0", "beta/1", "beta/3", "beta/39", "beta/4", "beta/40", "beta/41", "beta/42", "beta/43", "beta/44" ]
```

Checking dashboard.js ... there's a `getHumanReadableOptions` that filters out versions based on having 1 separate each channel. That should be in `telemetry.js` if it is a known problem with the API. I can filter it in wrapper, but then I'd have to filter it again in dashgen and again in the e10s dash... and anywhere `Telemetry.getVersions` is called.

Would pings from versions closer than 1 ever be legitimate? If not, I can fix telemetry.js.

>Additionally that generator link above never finishes loading

The generator link is broken because I misordered an if statement in telemetry-wrapper and throw during a JSON.parse of the `filters` empty string. I'll fix that.

>Use medians instead of means 

Sure, I'll just replace `.means()` with `.percentiles(50)`

t.m.o's evo display has... _interesting_ transforms for what the percentiles mean: https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/new-pipeline/src/evo.js#L7 For instance, take a gander at [this plot](http://mzl.la/1NeheTP) of beta firspaint numbers. I guess I should fix those while I'm in the neighbourhood?

>I think people will want a provision for custom titles, but let's cross that bridge when we get to it

Indeed. Once we allow that, we'll need to allow all sorts of customization. Right now, I hope providing some example custom CSS will encourage users to use CSS to customize. Don't want the default title or subtitle? `display: none` Want different line colours? Override the colors from `telemetry-wrapper.css`.

chutten:
More on versions. Apparently even aggregates.t.m.o doesn't know what correct versions are. Some misbehaving users could report telemetry from "nightly/99" and t.m.o will suffer an ignoble death (well, it wouldn't be happy).

The only definitive source for version information @mreid-moz and I could seem to find after long seconds of Googling was [in the source tree](https://hg.mozilla.org/mozilla-central/raw-file/default/config/milestone.txt). 

Which telemetry.js could totally request and parse to get the latest nightly, which would get it the latest everything else by process of decrementation... but it seems a little suspect to just nab data from an unrelated location in the mozilla-central tree and embed knowledge of the train release model into a client application.

Also I'm not sure at what point in the telemetry server code the version verification should take place. Or if it shouldn't at all and instead force that responsibility on all the clients.

So, for now I'll change telemetry-wrapper and dashboard-generator to use the t.m.o approach of "subtract by one and hope the latest nightly is valid" and take the larger "Telemetry and Versions" question to [the mailing list](http://mzl.la/1lk9PaS).

mreid-moz:
We probably wouldn't want to hammer the mercurial server to fetch that file too much either :-/

chutten:
That ought to take care of the rest of the concerns. Demo's been updated: https://people.mozilla.org/%7Echutten/telemetry-dashboard/dashboard-generator/index.html

vdjeric:
You should filter out Windows_95 and Windows_98 from the OS list. This was a bug in reporting that got fixed over the summer but I guess some old builds are still polluting the data. 

I expected there are other smaller bugs like this, but I think I'll merge this prototype (after the above bug is fixed) and let a couple of teams test it out.

chutten:
As before, the whole thing should be up-to-date on my p.m.o page: http://people.mozilla.org/~chutten/telemetry-dashboard/

Issue #130, opened 2015-08-18
"Compare by" for boolean histograms shouldn't show a 3-point line


vdjeric:
These should be bar graphs or multiple pie charts or something else, definitely not a 3-point curve http://bit.ly/1NDvfIn

chutten:
I think a grouped vertical bar graph would be ideal, but MG doesn't support it. I have filed [an issue](https://github.com/mozilla/metrics-graphics/issues/554) on MG to see if it can be added.

As for multiple pies: with the current design it wouldn't be simple to commandeer the selected-key titles to become compared measures. And if the number of compared measures exceeded four, there's no place to put the extra charts.

chutten:
@vdjeric, how about something like this? http://people.mozilla.org/~chutten/media/barGroup.PNG 

Since there's only really one independent measurement (False is 100% - True), I can munge the histograms into a single histogram showing multiple compare values' values. A few tweaks were needed here and there to make the (often long) labels display nicely.

vdjeric:
Yup that would work

PR #200, opened 2015-12-01
dash+wrapper: Add a "grouped bar" chart for comparing multiple boolean/flag histograms

chutten:
Looks like this: http://people.mozilla.org/~chutten/media/barGroup.PNG

Live version here: http://bit.ly/1QRXUya (the dashgen doesn't work live on that site because the generated dashes all point back to t.m.o)

Covers #130


vdjeric:
Looks good. I noticed some bugs, but they're not new:
* Linux is missing from the compare-by-OS-version view
* Windows_95 is showing up in the OS comparison view

PR #204, opened 2015-12-04
Update dashboards to point to data produced in new-dev IAM.

vitillo:
See fhr-dev [thread](https://mail.mozilla.org/pipermail/fhr-dev/2015-November/000672.html).


mreid-moz:
We should change the urls to use `analysis-output.telemetry.mozilla.org` instead of the S3 bucket URL directly.

Issue #212, opened 2016-01-11
Use % symbol on boolean evo graphs legends


vdjeric:
It's confusing to see figures such as "0.42" on a graph of percentages.. it's unclear whether it represents 42% or 0.42%.. hence we should use the percentage symbol for the tooltip legend

http://mzl.la/1ZX84jb

PR #213, opened 2016-01-12
dash: The mouseover text isn't using % signs for boolean plots.

chutten:
It thought that only enumerated plots needed % signs.

Issue #212


Uberi:
Looks good to me :+1:

Issue #214, opened 2016-01-12
Pie Charts list 'samples' when it means percentages.


chutten:
Must've messed up the d3pie call.

PR #215, opened 2016-01-12
wrapper: d3pie doesn't need precomputed %ges

chutten:
It can work just fine with counts. For the mouseover, it actually needs them.

Issue #214


Uberi:
Cool beans :+1:

Issue #206, opened 2015-12-09
Sanitized compare histograms show incorrect labels


chutten:
http://mzl.la/1OTRWto (for instance) has Windows_95 where the same dashboard as generated by http://mzl.la/1OTS506 calls it, correctly, Darwin.

This is probably due to sanitization not sanitizing out the correct label. I had this bug before launching dashgen.

PR #216, opened 2016-01-14
dash: If the data's been sanitized away, don't use its label.

chutten:
Issue #206


Uberi:
Looks good to me :+1:

PR #220, opened 2016-02-02
Fix "uncaught exception: Kind must be specified"

Uberi:
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1245114

The assertion is not necessary since we don't really need the histogram kind at that point. When there are no dates with available aggregate data, we should simply display a blank result rather than giving an assertion error.


chutten:
Seems legit

PR #227, opened 2016-03-08
Fix mixed content warning.

sapohl:
@chutten - Sorry, I didn't notice this mixed content warning on http://sapohl.github.io/telemetry-dashboard/update-orphaning/ because I didn't check the https:// variant of the URL, but this warning breaks https://telemetry.mozilla.org/update-orphaning/ since it's only served via https://.

### DIFF
--- a/update-orphaning/index.html
+++ b/update-orphaning/index.html
@@ -2,8 +2,8 @@
 <html>
 <head>
 <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


chutten:
For future reference, you can use urls that begin with `://` and it'll use the same protocol as the main content load.

Issue #229, opened 2016-03-24
mzl.la links should be https://


martinthomson:
No reason not to as far as I can see.

digitarald:
Fixed in #230

PR #230, opened 2016-03-24
Use https for shortened links.

chutten:
bit.ly seems to return only http links, so let's detect and fix that.

Issue #229

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-dashboard/230)


mreid-moz:
lgtm

PR #236, opened 2016-07-05
Add a page title and version distribution pie chart to the update orphaning dashboard

sapohl:
@chutten, if you could review/merge or forward, I'd appreciate it. Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-dashboard/236)

### DIFF
--- a/update-orphaning/index.html
+++ b/update-orphaning/index.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
+<title>Update orphaning dashboard</title>
 <meta charset="utf-8">
 <script type="text/javascript" src="https://ajax.googleapis.com/ajax/libs/jquery/1.6.2/jquery.min.js"> </script>
 <script type="text/javascript" src="https://d3js.org/d3.v3.min.js" charset="utf-8"></script>
--- a/update-orphaning/index.html
+++ b/update-orphaning/index.html
@@ -181,24 +181,6 @@ The Spark script for the weekly reports is scheduled to run every Wednesday and 
       var updateDisabled = data[1]["update-enabled-disabled"]["update-disabled"];
       var first = true;
 
-      // Generates the slices for a d3pie chart.
-      // @param aSlicesArray: an array of arrays in the following format:
-      //                      [[label, value, color],
-      //                       [label, value, color],
-      //                       ...
-      //                      ]
-      // @return An array of d3pie slices that can be used for data.content
-      function generateD3PieSlices(aSlicesArray) {
-        var slices = [];
-        for (s in aSlicesArray) {
-          var slice = { "label": aSlicesArray[s][0],
-                        "value": aSlicesArray[s][1],
-                        "color": aSlicesArray[s][2] };
-          slices.push(slice);
-        }
-        return slices;
-      }
-
       // Display a d3pie chart.
       // @param aDivId: a div id to populate with the d3pie
       // @param aTitle: a title for the d3pie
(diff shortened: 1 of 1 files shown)


update-orphaning/index.html
```diff
@@ -424,6 +430,117 @@
...
+            "46": "#99004C",
+            "47": "#A0A0A0",
+            "48": "#CCCC00",
+            "49": "#000066",
+            "50": "#0000FF",
+            "51": "#9933FF",
+            "52": "#009900",
+            "53": "#FFE5CC",
+            "54": "#663300",
+            "55": "#006666",
```
chutten: This seems like something that could get out of hand, maintenance-wise.

Perhaps assign specific colours based on number of versions behind current?

If you're looking for colour-blind-friendly colours, you can check out the work I did for telemetry wrapper: https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/wrapper/telemetry-wrapper.css#L85

update-orphaning/index.html
```diff
@@ -424,6 +430,117 @@
...
+          }
+          for (v in orphanedVersions) {
+            var color = "#0255d1";
+            if (v.split(".") && v.split(".")[0] &&
```
chutten: FWIW, `aString.split(delimeter)` will always return an array with at least one element

sapohl:
Thank you for the feedback! I used this opportunity to refactor the code to generate d3pie charts to avoid duplicating a bunch of code every time we want to display a new chart in the dashboard.

update-orphaning/index.html
```diff
@@ -181,17 +181,39 @@
...
+      // Generates the slices for a d3pie chart.
+      // @param aSlicesArray: an array of arrays in the following format:
+      //                      [[label, value, color],
+      //                       [label, value, color],
+      //                       ...
+      //                      ]
+      // @return An array of d3pie slices that can be used for data.content
+      function generateD3PieSlices(aSlicesArray) {
+        var slices = [];
+        for (s in aSlicesArray) {
```
chutten: I think you mean to use for..of, not for..in (which will iterate over length as well :(

And you can be clever, if you'd like, and use [destructuring assignment](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) like so:

```javascript
for ([label, value, colour] of aSlicesArray) {
  slices.push({label: label, value: value, color: colour});
}
```

...actually, it might be cleaner still to
```javascript
return aSlicesArray.map(([label, value, colour]) => {
  return {"label": label, "value": value, "color": colour};
});
```

update-orphaning/index.html
```diff
@@ -430,120 +448,55 @@
...
             }
-            var slice = {
-                          "label": v,
-                          "value": orphanedVersions[v],
-                          "color": color
-                        };
-            versionDistributionObject.push(slice);
+            var color = versionColors[index];
+            var slice = [v, orphanedVersions[v], color];
+            versionSlices.push(slice);
```
chutten: Since you're generating slices here, I see no reason not to build them straightaway into d3 format

update-orphaning/index.html
```diff
@@ -430,120 +448,55 @@
...
+            if (diff < versionColors.length) {
+              index = diff;
             }
-            var slice = {
-                          "label": v,
-                          "value": orphanedVersions[v],
-                          "color": color
-                        };
-            versionDistributionObject.push(slice);
+            var color = versionColors[index];
```
chutten: I think combining some lines may aid comprehension:

```javascript
var diff = latestVersion - v.split(".")[0];
var color = versionColors[Math.min(diff, versionColors.length - 1)];
```

PR #253, opened 2016-10-24
Bug 1312420 - Don't try to use + on things that could be either strings or numbers

matt-howell:
This dashboard builds a file name to fetch by using + to concatenate a year, month, and day. If the month is before October, it has to get a 0 prepended, forcing its conversion to a string. But in October or later, when that doesn't happen, the components are all still numbers and + adds them up as numbers instead of concatenating them as strings.
So, here we switch to a method of building the file name string that doesn't care what the types of the components are.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-dashboard/253)


matt-howell:
needs-review: @chutten

chutten:
LGTM

PR #266, opened 2016-11-25
Timezone Fixes

chutten:
It turns out that we've been showing times wrong on the evolution view for some... er, time. It all depended what timezone you were viewing the page from (of course) so that might explain why it's flown under the radar for so long.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-dashboard/266)

### DIFF
--- a/new-pipeline/src/evo.js
+++ b/new-pipeline/src/evo.js
@@ -799,6 +799,7 @@ function displayEvolutions(lines, submissionLines, useSubmissionDate,
     legend: aggregateLabels,
     aggregate_rollover: true,
     linked: true,
+    utc_time: true,
     mouseover: function (d, i) {
       var date, rolloverCircle, lineList, values;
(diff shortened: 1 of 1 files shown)


fbertsch:
:lgtm:

Only thought might be, should we put somewhere that says "All times in UTC"? We had to add something like that to ATMO. Maybe mg.js does it for us?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-dashboard/266#-:-KXkzED9YQPkXFfSzdrR:b-usargt)*

chutten:
The popup legends all state UTC explicitly. They were just lying about it before :)

fbertsch:
Heh, of course. Quality patch then :)

PR #268, opened 2016-12-15
Remove unused/broken dashboards

chutten:
There are likely others, but here are a few obvious choices.

r? @bsmedberg

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-dashboard/268)


ddurst:
I think we should give them an escape route instead of just removing the entry. 
page-shot-2016-12-19 telemetry dashboards
(that content is FPO, but you get the gist)

ddurst:
wfm. r+ (now I gotta set an alert on issues...)

Issue #281, opened 2017-01-31
Stability Dashboard should use latest released version, not most popular


chutten:
[Presently](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/crashes/index.html#L106) the code checks for the most popular release by number of kilo usage hours.

It would be far more relevant to relman and uptime to see the figures for the latest release, even if it does cause the kuh plot to dip on merge day.

ddurst:
Can we just toggle between the two?

chutten:
Why? Do you know of anyone who would prefer to look at an older-than-current version?

ddurst:
Nevermind! I misread this.

chutten:
Live preview can be found here: https://chutten.github.io/telemetry-dashboard/crashes/

PR #282, opened 2017-02-01
Issue #281 - Use the most-recent version, not the most-used version for the crash dash

chutten:
The crash dash currently displays the most-used version per channel in an
effort to maximize the crash rate denominator "kuh".

This isn't terribly helpful as, once a new release lands, we cease to really
care about the stability of the old one.

So track the release schedule using https://product-details.mozilla.org and
only use data from the most recent release.

Issue #281

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-dashboard/282)

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -44,7 +44,7 @@
                     with
                     <select id="filter-e10s" class="multiselect" title="e10s setting" data-all-selected="any e10s setting" multiple></select>
                     as
-                    <select id="filter-process-type" class="multiselect" title="Process type" data-all-selected="any process type" multiple></select>
+                    <select id="filter-process-type" class="multiselect" title="Process type" data-all-selected="any process type"></select>
                     and compare by
                     <select id="compare" class="multiselect" title="Filter name to compare options for">
                         <option value="">none</option>
--- a/new-pipeline/evo.html
+++ b/new-pipeline/evo.html
@@ -54,7 +54,7 @@
                     with
                     <select id="filter-e10s" class="multiselect" title="e10s setting" data-all-selected="any e10s setting" multiple></select>
                     as
-                    <select id="filter-process-type" class="multiselect" title="Process type" data-all-selected="any process type" multiple></select>
+                    <select id="filter-process-type" class="multiselect" title="Process type" data-all-selected="any process type"></select>
                 </h2>
                 <h3 id="measure-description" style="font-size: 14px; margin: 0 3px"></h3>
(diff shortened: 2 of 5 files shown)


chutten:
Live preview can be found here: https://chutten.github.io/telemetry-dashboard/crashes/

chutten:
This needs some tweaking. It needs to auto-select a day from the latest version instead of waiting for usage to increase enough to match the previous version's use, and the crash plot can probably just stop truncating on the x-axis.

crashes/latest-versions.js
```diff
@@ -0,0 +1,86 @@
...
+
+  let addVersion = function (version, add) {
+    if (!add) {
+      return version;
+    }
+    let major = version.split('.')[0];
+    return (window.parseInt(major, 10) + add) + ".0"
+  };
+
+	var formatYYYYmmdd = function(d) {
```
fbertsch: Indentation?

crashes/index.html
```diff
@@ -61,6 +62,9 @@
 var gData = undefined;
 var gMap = {};
 
+// When populated, this is a map activity_date -> channel -> version
+var gLatestVersions = getLatestVersions();
```
fbertsch: With limit=10 (at least, I *think* that is what is causing the problem), some options for "analyzed date" aren't showing (well, they show, but they don't update the table). Check the earliest date (2016-11-08) for an example of this. Gets a "TypeError: gMap[weekAgoDate] is undefined".

fbertsch (commented):
1. The y-axis is *way* below the max y-value, is this expected?

2. The % next to each number doesn't seem very relevant anymore (sways wildly between 10^-2 and 10^4, depending on how far after the release). Maybe it's best to compare the X days after the previous release for that channel?

crashes/index.html
chutten: The first seven dates don't work as they rely on data being present a week before. I suppose I can make that safe, but it's always been that way. I've filed #284 for this.

chutten (commented):
1. The y axis is locked so that crash spikes don't reduce the rest of the graph to a flat line near 0.

2. I've struggled with the usefulness of the %ge stuff. It's a real pain, but if I remove them users may misunderstand the veracity of the dash's claims. As I can't guarantee that the data for multiple releases will be the same (or even present in the query results) I don't think I can use "release + x days" as a metric. (especially if a release is delayed).

This leaves me unfortunately uncertain about what to do with them.

fbertsch (approved):
LGTM

PR #295, opened 2017-03-02
Harden the dashboard against crash types that may not have thresholds

chutten:
Updating the query and the thresholds.js can't happen simultaneously, so don't assume that we have thresholds for all kinds of crashes.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-dashboard/295)


fbertsch (approved):
LGTM

Issue #291, opened 2017-02-27
factor in gpu-process crashes into the crash dashboard


philipp-sumo:
firefox 53 introduces a new gpu process ("quantum compositor") - we should find a way to incorporate that into the crash dashboard. 

at the moment it looks like in socorro there is a rather small number of those reports, so maybe we can just count those crashes into the content crash rate (also see issue #290), as those types of crashes might feel similar to affected users, and add a note about this on the dashboard.

chutten:
The dashboard as it stands is set to show rates instead of counts, so it'll require a little more effort to treat gpu crashes as a count. Shouldn't be too terrible.

That being said, I'm not sure that counts give us anything better than rates. So long as the proportion of the population getting gpu processes (and, thus, gpu crashes) is a representative %ge of the overall population, then it'll likely still follow the same weekly/seasonal ups and downs that the "per-kilo-usage-hour" rate was designed to normalize.

I'll experiment a bit here and see what it looks like.

philipp-sumo:
sorry if i haven't expressed myself clearly. i'm not interested in a count per se, but just wanted to raise the issue that gpu process crashes should be factored into the rates "somehow". i think they could be treated as part of the content crash rates for example...

chutten:
No, no, it's fine. We each have our own vocabulary with our own meanings for words. For me a "count" is not useful un-normalized (I've actually [blogged on the subject](https://chuttenblog.wordpress.com/?p=2493)) and is distinct from a "rate".

I can add gpu_crashes per thousand usage hours as a row with little problems, once I clear up some [pending issues](https://github.com/mozilla/telemetry-dashboard/pull/282) first.

chutten:
Give it a go now? How's it look?

philipp-sumo:
thank you!

Issue #290, opened 2017-02-27
Remove "M+C" from the dashboard in favor of "C" or "C-S"


philipp-sumo:
this is an enhancement request: 
the telemetry-fed crash dashboard at https://telemetry.mozilla.org/crashes/ lists a "Main (M)" crash rate for crashes in the browser process. it would be helpful if we could have a dedicated rate that shows the ups/down just for content crashes as well.

chutten:
Give it a go now? How's it look?

philipp-sumo:
looking great, thank you!

PR #296, opened 2017-03-02
Small fixes

chutten:
Fix an error that stopped the y-axis clamping from working

Add in thresholds for one of the two new crash types. There's no possible way to threshold GPU at the moment, so leave it out of it. (this means it also doesn't benefit from y-axis clamping until we set the thresholds to something).

It amounts to little more than spit and polish.

Finishes work on #291 and #290 begun with #295 (and some changes within sql.tmo)

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-dashboard/296)

### DIFF
--- a/crashes/index.html
+++ b/crashes/index.html
@@ -349,7 +349,6 @@ function updateCrashPlot(date, channel, crash) {
     }
   });
   let baselines = [];
-  maxY = undefined;
   if (crash in THRESHOLDS[channel]) {
     baselines = [
       {value: THRESHOLDS[channel][crash].lo, label: ""},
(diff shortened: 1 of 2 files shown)


crashes/index.html
```diff
@@ -349,7 +349,6 @@
     }
   });
   let baselines = [];
-  maxY = undefined;
```
fbertsch, line 352: Hmm, I don't understand this. Why doesn't it error on the MG.data_graphic now that maxY isn't defined at all? (for the crash types not in THRESHOLD[channel]

fbertsch (changes requested):

fbertsch (approved):
stupid js.

PR #301, opened 2017-03-22
bug 1349536 - Add "Other" channel to crashdash

chutten:
With users of old versions of Windows having been migrated to ESR in 52, it
suddenly becomes interesting to look at ESR crash rates and adoption.

Sadly ESR doesn't fit into the nice "latest versions" mechanism, so for now it
is reporting the most-used version. This works well for the latest ESR because
of the aforementioned influx of users, but may need to be tweaked in time for
the next one.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-dashboard/301)

### DIFF
--- a/crashes/index.html
+++ b/crashes/index.html
@@ -28,6 +28,7 @@
       <td>Aurora</td>
       <td>Beta</td>
       <td>Release</td>
+      <td>Other</td>
     </tr>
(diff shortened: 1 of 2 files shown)


fbertsch (approved):
LGTM

PR #309, opened 2017-05-04
Bug 1299490 added Submissions to filter

aearl16:
@chutten - Changed line 230, added submissions to the list of strings.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-dashboard/309)

### DIFF
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -229,7 +229,7 @@ function loadStateFromUrlAndCookie() {
     var aggregates = pageState.aggregates.split("!")
       .filter(function (v) {
         return ["5th-percentile", "25th-percentile", "median",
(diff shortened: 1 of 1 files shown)


chutten:
Have you tested it? Does it work?

aearl16:
@chutten - I believe so. We got the 'submissions' in the drop down list when we opened the page.

chutten:
@aearl16 That's a good first step. The more direct test is to, after the page completely loads, select "submissions" from the dropdown so it populates the plot. Then copy the URL into a fresh tab and load it.

The new tab must have the submissions line on the plot to pass the test.

aearl16:
We did do that, but when the page is opened and we clicked submissions the page says the data is no longer available.

chutten:
That's to be expected. If you unselect all of the aggregates, it has nothing to show. Thank you!

aearl16:
Ya no problem.

PR #314, opened 2017-05-22
Allow recent release and beta versions

fbertsch:
You can test this at https://fbertsch.github.io/telemetry-dashboard/

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-dashboard/314)

### DIFF
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -669,10 +669,10 @@ function getHumanReadableOptions(filterName, options) {
           (version <= latestNightlyVersion - 1 ? goodOptions : badOptions)
           .push(option);
         } else if (parts[0] === "beta") {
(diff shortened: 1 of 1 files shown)


new-pipeline/src/dashboards.js
```diff
@@ -669,10 +669,10 @@ function getHumanReadableOptions(filterName, options) {
           (version <= latestNightlyVersion - 1 ? goodOptions : badOptions)
```
chutten, line 669: Should we also instead put all aurora > 53 in the badlist, and nix it from the "latest in each channel" bit, below?

Or is that a story for another time?

fbertsch: Looks like Aurora is actually sticking around, which makes our life easier: https://bugzilla.mozilla.org/show_bug.cgi?id=1362137#c1

chutten (approved):

PR #327, opened 2017-08-24
Fix the recommendation page to use the new mapping format

Dexterp37:
The format changed from {"hashed_id": "name"} to
{"hashed_id": {"name": "Some Name", "id": "addon guid", ...}.
This commit also makes the recommender only recommend
non-legacy addons.


maurodoglio (approved):
LGTM

PR #333, opened 2017-09-18
Bug 1365751 - correctly handle keyed histograms with a single key

adrian17:
Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1365751

Bug repro: [link](http://telemetry.mozilla.org/new-pipeline/dist.html#!cumulative=0&end_date=2017-08-25&keys=content%253Aamd64%252F49!__none__!__none__!__none__&max_channel_version=aurora%252F56&measure=SANDBOX_REJECTED_SYSCALLS&min_channel_version=null&os=Linux&processType=*&product=Firefox&sanitize=0&sort_keys=submissions&start_date=2017-08-21&table=0&trim=1&use_submission_date=0)
Same page, temporary server with this patch applied: [link](http://212.71.251.88:8050/new-pipeline/dist.html#!cumulative=0&end_date=2017-08-25&keys=content%253Aamd64%252F49!__none__!__none__!__none__&max_channel_version=aurora%252F56&measure=SANDBOX_REJECTED_SYSCALLS&min_channel_version=null&os=Linux&processType=*&product=Firefox&sanitize=0&sort_keys=submissions&start_date=2017-08-21&table=0&trim=1&use_submission_date=0)

Also changed the behavior for CSV/JSON export.

In a separate commit, fixed histogram caption not always appearing. I noticed this when testing the first commit.

### DIFF
--- a/new-pipeline/src/dist.js
+++ b/new-pipeline/src/dist.js
@@ -881,6 +881,13 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
     });
   }
 
+  if (histogramsList.length > 0 && histogramsList[0].histograms.length > 0) {
+    // Set the histogram caption to the histogram description
+    $('#dist-caption').text(histogramsList[0].histograms[0].description);
+  } else {
+    $('#dist-caption').text(""); // Clear the histogram caption
+  }
+
   if (histogramsList.length <= 1) { // Only one histograms set
     if (histogramsList.length === 1 && histogramsList[0].histograms.length ===
       1) { // Only show one set of axes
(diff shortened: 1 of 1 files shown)


chutten:
Is there a way that we could have a full-size keyed histogram display for the single-key case? Sort of like the non-keyed display case, but with a title?

adrian17:
I was considering suggesting doing it in a separate feature, but it turned out to be pretty easy. Updated the first commit and test server.

chutten:
Looks good! I will rebase your patches into the repo and then your changes should go live within the next ten minutes or so.

PR #335, opened 2017-10-01
Add `metric count` evolution plot; fixes #312.

### LINKED ISSUE #312: Evolution View should have a way to see "Total Histogram Count"
We currently have "Number of Pings" with the label "Daily Metric Count". I would prefer to have to plots with "counts":
1. "Daily Ping Count" - number of pings received in that day with that metric
2. "Daily Metric Count" - number of instances of that metric recieved per day.

adrian17:
Fixes #312.

Before: [link](https://telemetry.mozilla.org/new-pipeline/evo.html#!aggregates=median&cumulative=0&end_date=null&keys=&max_channel_version=nightly%252F58&measure=FX_TAB_SWITCH_UPDATE_MS&min_channel_version=nightly%252F55&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=null&trim=1&use_submission_date=0)
After: [link](https://adrian17.github.io/telemetry-dashboard/new-pipeline/evo.html#!aggregates=median&cumulative=0&end_date=null&keys=&max_channel_version=nightly%252F58&measure=FX_TAB_SWITCH_UPDATE_MS&min_channel_version=nightly%252F55&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=null&trim=1&use_submission_date=0)

Changes:
- the new plot is displayed under the the "submissions" plot.
- "submissions" option was removed from the aggregates multiselect. I didn't feel it's needed, since the submissions plot is always shown anyway
- a lot of refactors, just to make the actual feature implementation commit just 20 lines long

To consider:
- I'm 90% sure I got the "metric count" semantics right (it's equivalent to the `Sample Count` value in Distribution View), but it'd be nice if you made sure I did it right (see side question below)
- The new plot is almost identical to the submissions plot (with the exception of order of magnitude of values), which makes them look a bit weird. There may be a smarter way to draw these two, but I could figure out a good (and easy to implement) UX for that.
- is refactoring/reformatting/simplifying related code in feature PRs (see commit d634adef88718c2b121318874587730790d38764) like this OK, or is it too much code churn?

Side question: sometimes/often `metric count` < `ping count`. For example [distribution](https://telemetry.mozilla.org/new-pipeline/dist.html#!cumulative=0&end_date=2017-08-01&keys=__none__!__none__!__none__&max_channel_version=nightly%252F56&measure=BROWSER_IS_USER_DEFAULT&min_channel_version=nightly%252F55&os=Windows_NT%252C6.2&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=2017-06-12&table=0&trim=1&use_submission_date=0), [evolution](https://adrian17.github.io/telemetry-dashboard/new-pipeline/evo.html#!aggregates=bucket-0!bucket-1!bucket-2&cumulative=0&end_date=null&keys=&max_channel_version=nightly%252F58&measure=BROWSER_IS_USER_DEFAULT&min_channel_version=nightly%252F55&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=null&trim=1&use_submission_date=0). I don't get why exactly this can happen. Does the server report _all_ pings in the time range, even if they didn't have this value? (which also seems a bit strange, since 1. that would mean the `ping count` should be equal between metrics, but it isn't, and 2. being default browser is probably checked fairly early on?)

### DIFF
--- a/new-pipeline/src/evo.js
+++ b/new-pipeline/src/evo.js
@@ -735,20 +735,17 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
     }); // Need to add 1ms because the leftmost marker won't show up otherwise
   }
   if (markers.length > 1) { // If there is a marker on the far right, move it back 2 milliseconds in order to make it visible again
-    markers[markers.length - 1].date = moment.utc(markers[markers.length - 1].date.getTime() -
-        2)
-      .toDate();
+    let lastMarker = markers[markers.length - 1];
+    lastMarker.date = moment.utc(lastMarker.date.getTime() - 2).toDate();
   }
 
-  d3.select(`${target} .active-datapoint-background`)
-    .remove(); // Remove old background
+  d3.select(`${target} .active-datapoint-background`).remove(); // Remove old background
   MG.data_graphic({
     data: lineData,
     chart_type: lineData.length == 0 || lineData[0].length === 0 ?
       "missing-data" : "line",
-    width: $(target)
-      .parent()
-      .width(), // We can't use the full_width option of MetricsGraphics because that breaks page zooming for graphs
+    // We can't use the full_width option of MetricsGraphics because that breaks page zooming for graphs
+    width: $(target).parent().width(),
     height: 600,
     right: 100,
     bottom: 50, // Extra space on the right and bottom for labels
@@ -758,11 +755,7 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
     y_label: "",
     transition_on_update: false,
     interpolate: "linear",
-    yax_format: usePercentages ? function (y) {
-      return y + "%";
-    } : function (y) {
-      return y;
-    },
+    yax_format: usePercentages ? (y => y + "%") : null,
     markers: markers,
     aggregate_rollover: true,
     linked: true,
@@ -780,12 +773,8 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
           seen[entry.line_id] = true;
           return true;
         });
-        lineList = entries.map(function (entry) {
-          return lines[entry.line_id - 1];
-        });
-        values = entries.map(function (entry) {
-          return entry.value;
-        });
+        lineList = entries.map(entry => lines[entry.line_id - 1]);
+        values = entries.map(entry => entry.value);
       } else {
         date = d.date;
         rolloverCircle = $(`${target} .mg-line-rollover-circle`).get(0);
@@ -810,8 +799,7 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
             values[i]) + (usePercentages ? "%" : ""));
         legend.append("tspan")
           .attr({
-            x: -label.node()
-              .getComputedTextLength(),
+            x: -label.node().getComputedTextLength(),
             y: (lineIndex * lineHeight) + "em"
           })
           .text("\u2014 ")
@@ -825,11 +813,11 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
       var x = parseInt(rolloverCircle.getAttribute("cx")) + 20,
         y = 40;
       var bbox = legend[0][0].getBBox();
-      if (x + bbox.width + 50 > $(`${target} svg`)
-        .width()) x -= bbox.width + 40;
+      if (x + bbox.width + 50 > $(`${target} svg`).width()) {
+        x -= bbox.width + 40;
+      }
       d3.select(`${target} .mg-active-datapoint-container`)
-        .attr("transform", "translate(" + (x + bbox.width) + "," + (y +
-          15) + ")");
+        .attr("transform", `translate(${x + bbox.width}, ${y + 15})`);
 
       // Add background
       var padding = 10;
@@ -847,8 +835,7 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
         .style("fill", "#333");
     },
     mouseout: function (d, i) {
-      d3.select(`${target} .active-datapoint-background`)
-        .remove(); // Remove old background
+      d3.select(`${target} .active-datapoint-background`).remove(); // Remove old background
     },
     ...plotOptions
   });
@@ -895,18 +882,12 @@ function displayEvolutions(lines, submissionLines, useSubmissionDate,
   usePercentages) {
   indicate("Rendering evolutions...");
 
-  var aggregateLabels = lines.map(function (line) {
-    return line.aggregate;
-  });
-
-  var aggregateMap = {};
-  lines.forEach(function (line) {
-    aggregateMap[line.aggregate] = true;
-  });
+  var aggregateLabels = lines.map(line => line.aggregate);
+  var aggregateSet = new Set(aggregateLabels);
   var variableLabel = useSubmissionDate ?
     "Submission Date (click to use Build ID)" :
     "Build ID (click to use Submission Date)";
-  var valueLabel = Object.keys(aggregateMap)
+  var valueLabel = [...aggregateSet]
     .sort()
     .join(", ") + " " + (lines.length > 0 ? lines[0].measure : "");
 
--- a/new-pipeline/src/evo.js
+++ b/new-pipeline/src/evo.js
@@ -33,6 +33,9 @@ var gMetaAggregates = [
   ["submissions", "Submissions", function (evolution) {
     return evolution.submissions();
   }, "#submissions"],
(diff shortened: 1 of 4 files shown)


chutten:
The live link looks great!

Before I get into the code: generally big refactors are less useful that they appear. There's a lot to be gained by treating every unfamiliar line of code as the fix to a bug we didn't even see. That being said, these dashboards were coded rapidly, so there's probably some easy wins a refactor could provide :)

To your side question: "ping count" is only counting pings that contain the probe. "sample count" is the number of measurements. So two pings with histogram `A`, one with one measurement of '1' and the other with two measurements, of '1' and '7' would have a "ping count" of 2, a "sample count" of 3, and a "sample sum" of 9.

Those links you provide are likely a bug. Thanks for catching them! Thanks to your report @fbertsch is now looking into it.

new-pipeline/src/evo.js
```diff
@@ -758,11 +757,7 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
     y_label: "",
     transition_on_update: false,
     interpolate: "linear",
-    yax_format: usePercentages ? function (y) {
-      return y + "%";
-    } : function (y) {
-      return y;
-    },
+    yax_format: (y => y + valueSuffux),
```
chutten: This coerces y to a string which, for counts like submissions and metric counts, means the formatter can't automatically translate the y-ticks from, say, 150000 to 150k.

new-pipeline/src/evo.js
```diff
@@ -703,6 +703,8 @@ function getHistogramEvolutionLines(channel, version, measure, aggregates,
 }
 
 function displayEvolution(target, lines, usePercentages, plotOptions) {
+  let valueSuffux = usePercentages ? "%" : "";
```
chutten: s/Suffux/Suffix/g

new-pipeline/src/evo.js
```diff
@@ -825,11 +814,9 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
       var x = parseInt(rolloverCircle.getAttribute("cx")) + 20,
         y = 40;
       var bbox = legend[0][0].getBBox();
-      if (x + bbox.width + 50 > $(`${target} svg`)
-        .width()) x -= bbox.width + 40;
+      if (x + bbox.width + 50 > $(`${target} svg`).width()) x -= bbox.width + 40;
```
chutten: The style we're aspiring to use recommends three-line conditionals
```
if (cond) {
  ...body...
}
```

new-pipeline/src/evo.js
```diff
@@ -825,11 +814,9 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
...
       d3.select(`${target} .mg-active-datapoint-container`)
-        .attr("transform", "translate(" + (x + bbox.width) + "," + (y +
-          15) + ")");
+        .attr("transform", "translate(" + (x + bbox.width) + "," + (y + 15) + ")");
```
chutten: More concise as a template string?
`translate(${x + bbox.width}, ${y + 15})`

new-pipeline/src/evo.js
```diff
@@ -25,10 +24,14 @@ var gDefaultAggregates = [
   ["95th-percentile", "95th percentile", function (evolution) {
     return evolution.percentiles(95);
   }],
+];
+// these will not appear in the multiselect
```
chutten: I'm not sure if anyone uses it, but being able to have the submissions on the same plot instead of just on the linked plot below might support use cases requiring a more closer scrutiny of submissions vs values. For now I recommend leaving it in the multiselect.

new-pipeline/src/evo.js
```diff
@@ -877,10 +859,14 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
 
 }
 
-function displayEvolutions(lines, submissionLines, useSubmissionDate,
-  usePercentages) {
+function displayEvolutions(lines, useSubmissionDate, usePercentages) {
   indicate("Rendering evolutions...");
 
+  var submissionLines = lines.filter(line => line.aggregate == "Submissions");
```
chutten: I'm not the biggest fan of magic strings. Can we at least pull this out to a constant? Or if you can think of something more clever, I'd like that too :)

new-pipeline/src/evo.js
```diff
@@ -877,10 +859,14 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
...
+
+  let metaAggregateNames = gMetaAggregates.map(entry => entry[1]);
+  lines = lines.filter(line => !metaAggregateNames.includes(line.aggregate));
```
chutten: This imposes a constraint that aggregates and metaAggregates cannot share the same name. Not likely to come up, but you should put it in a comment up near gMetaAggregates.

new-pipeline/src/evo.js
```diff
@@ -887,6 +890,11 @@ function displayEvolutions(lines, useSubmissionDate, usePercentages) {
     x_label: variableLabel,
     y_label: "Daily Ping Count",
   });
+  displayEvolution('#metric-count', countLines, false, {
+    height: 300,
+    x_label: variableLabel,
+    y_label: "Daily Metric Coun",
```
chutten: missing a "t"

v2/telemetry.js
```diff
@@ -312,6 +312,10 @@
       });
     };
 
+    Evolution.prototype.metric_count = function () {
```
chutten: telemetry.js uses `camelCase` for its API names.

Speaking of which, if you're adding to the API you need to add to the documentation. telemetry.js is used in a few places by a few teams and individuals. Keeping the documentation up-to-date is of utmost importance.

chutten:
We've found out the problem that was showing you ping count > sample count. It's an obscure Telemetry client bug that you need not concern yourself about the details... the closest tl;dr I can give is: we were sending blank histograms for every ping after the first one for that probe (and others like it) which resulted in higher ping counts than sample counts.

[The bug](https://reviewboard.mozilla.org/r/154214/diff/4#index_header) was noticed and fixed as a part of a [large refactor](https://bugzilla.mozilla.org/show_bug.cgi?id=1366294) in the middle of Nightly 56 which is why data from 57 and 58 do not exhibit this character.

Thanks again for catching it!

adrian17:
Pushed updated commits.

After discussion on IRC, dropped [this comment](https://github.com/mozilla/telemetry-dashboard/pull/335#discussion_r142142153).

Renamed "metric count" to "sample count", as "sample" seemed to already be used in the UI and docs.
(Another note: I called the new Evolution API function `sampleCounts()`. While the naming is inconsistent with `submissions()`, I think it's a lesser evil -  `.submits()` would be a misleading name.)

chutten:
Is the new code being served properly at the "After" link? I keep refreshing and the lower graphs keep having y-axis ticks that are full of 0s instead of being automagically shortened to things like 150k...

adrian17:
Oh, turns out the old code had a small bug too.

> This coerces y to a string which, for counts like submissions and metric counts, means the formatter can't automatically translate the y-ticks from, say, 150000 to 150k.

Turns out even just `y => y` disables pretty formatting in MetricsGraphics.

Fixed it [here](https://github.com/mozilla/telemetry-dashboard/pull/335/commits/b8e3ba3f94aff48d7d891e5c44d957123aa3bfab#diff-ba801d8489368264e675c56b6b5d0817R758).
This also enables value formatting for the main plot - see [before](https://telemetry.mozilla.org/new-pipeline/evo.html#!aggregates=median!mean&cumulative=0&end_date=null&keys=&max_channel_version=nightly%252F58&measure=SEARCH_SERVICE_COUNTRY_FETCH_TIME_MS&min_channel_version=nightly%252F55&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=null&trim=1&use_submission_date=0) [after](https://adrian17.github.io/telemetry-dashboard/new-pipeline/evo.html#!aggregates=median!mean&cumulative=0&end_date=null&keys=&max_channel_version=nightly%252F58&measure=SEARCH_SERVICE_COUNTRY_FETCH_TIME_MS&min_channel_version=nightly%252F55&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=null&trim=1&use_submission_date=0).

chutten (approved):

chutten:
Looks good to me! Let's get this in production. (It should show up within the next half hour. Possibly faster)

PR #340, opened 2017-10-11
Issue #339 - Fix telemetry-wrapper to support D3 v4

armenzg:
@chutten r?

This works (only tested evoTime) with the project having metrics-graphics installed as a package and without.

Would you be able to test drawAsLines for me? (since you're more familiar on to quickly test it)

How long after merging does this new version get deployed?

### DIFF
--- a/wrapper/telemetry-wrapper.js
+++ b/wrapper/telemetry-wrapper.js
@@ -394,7 +394,7 @@ window.TelemetryWrapper.go = function (params, element) {
         (date.getMonth() + 1) + '-' + date.getDate(), // TODO: i18n
       yax_format: y => valuesArePercent ? y + '%' : y,
       transition_on_update: false,
-      interpolate: 'linear',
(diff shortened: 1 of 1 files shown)


chutten:
The current Readme for wrapper suggests using d3 v3.5.5 

Does the code as written work with that version? If not, it will need to as it might be live-linked from places that have the d3 version hard-coded.

armenzg:
I'm using that version via script:
https://github.com/armenzg/react-telemetry-demo/blob/master/public/index.html#L15

If I comment out this line:
https://github.com/armenzg/react-telemetry-demo/blob/metrics_graphics/src/App.js#L2-L5
it then uses what is specified in the <script> tag.

I've tested both case scenarios.
Feel free to test the "metrics_graphics" branch from that repo if my answer did not clarify it enough.

chutten:
I've given it a whirl and it works for me

PR #341, opened 2017-11-02
bug 1411305 - Notify users that the unrepresentative release data's going away.

chutten:
'cause it is. Soonish.


chutten:
Live preview: http://chutten.github.io/telemetry-dashboard/

fbertsch (approved):
🚢

georgf:
This is a bit out of context on the TMO landing page, i can't tell what this actually affects without reading the full blog post.
Can we make clear this only affects aggregates and the distribution/evolution dashboards?
Is it sufficient to show this notice on the distribution/evolution dashboard?

chutten:
I'm fine with removing it from the index.

PR #342, opened 2017-11-06
Remove "release aggregates going away" notice from index

chutten:
It's not 100% clear that it's only relevant to dist and evo.

### DIFF
--- a/index.html
+++ b/index.html
@@ -26,10 +26,10 @@
                 <a href="http://docs.telemetry.mozilla.org" class="btn" title="Docs" target="_blank">Telemetry Documentation</a>
             </form>
             <h2>Telemetry dashboards</h2>
-            <div class="error-msg">
+            <!-- <div class="error-msg">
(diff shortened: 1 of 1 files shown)


fbertsch (changes requested):
This commit should just be a revert of the previous one. We can reapply the previous one if we want to reuse any of its changes.

chutten:
@fbertsch A partial revert, you mean? Why would that be better than just commenting it out?

georgf (commented):
This part makes sense to me. The message will still be shown in the measurements & evolution dashboard.

Could we make it a little more focused on the impact though in the other places?

Maybe something like:
"The data from release users will go away soon, affecting this dashboard and the aggregates service. For context, please read ..."

fbertsch (approved):
Ah, I see now that it is still showing on the other views. LGTM :)

chutten:
I'll take the rewording request to a different PR

PR #345, opened 2017-11-22
Remove only target="_blank" usage

georgf:
This is the only place that is using target=“_blank”.
Erik caught that this is inconsistent and should probably just have
standard behaviour.

### DIFF
--- a/index.html
+++ b/index.html
@@ -17,7 +17,7 @@
     <div class="container-fluid">
         <header>
             <form class="navbar-form inline pull-right">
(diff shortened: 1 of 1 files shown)


chutten (changes requested):
I'm not 100% on board with this. I think if you click on report a bug, it usually means you are in a buggy state. With target=_blank the buggy state remains open and present to be inspected for writing the bug report.

fbertsch:
Is anyone really going to be in a buggy state on the TMO homepage? Also consider that then the behavior of this "Report Bug" and the "Report Bug" on the Measurement and Evolution view pages will behave differently.

chutten (approved):
Oh, good point. Bugs are far more likely in the dashboards themselves.

PR #352, opened 2017-12-05
bug 1293750 - Add Google Analytics to telemetry.mozilla.org

chutten:
We don't know where users come from or go to on TMO, and that's an odd sort of
lapse for a data-based web application.

This is preliminary: just gtags for now.

### DIFF
--- a/addon-perf/index.html
+++ b/addon-perf/index.html
@@ -12,14 +12,8 @@
     <script type="text/javascript" src="/lib/jquery.csv-0.8.2.js"></script>
     <link rel="stylesheet" href="addon-perf.css" />
     <link rel="stylesheet" href="http://code.jquery.com/ui/1.10.4/themes/smoothness/jquery-ui.css">
-    <script>
-      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-      (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-      m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-      })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
-      ga('create', 'UA-49796218-2', 'mozilla.org');
-      ga('send', 'pageview');
-    </script>
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-49796218-2"></script>
+    <script src="../analytics/analytics.js"></script>
 </head>
 <body>
 <!--
@@ -143,6 +137,7 @@
     </div>
 </div>
 <div id='footer'></div>
+<div><a href="https://www.mozilla.org/en-US/privacy/websites/">Mozilla's Website Privacy Notice</a></div>
 </body>
 </html>
 
--- a/analytics/analytics.js
+++ b/analytics/analytics.js
@@ -1,0 +1,29 @@
+(function() {
+
+window.trackingEnabled = () => {
+  var _dntStatus = navigator.doNotTrack || navigator.msDoNotTrack;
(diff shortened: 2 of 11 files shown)


chutten:
Interestingly we already had some analytics on the addon-perf page. It works, but I'm updating it here for consistency's sake.

Dexterp37:
Mh, is #353 doing the same?

addon-perf/index.html
```diff
@@ -12,13 +12,26 @@
...
     <link rel="stylesheet" href="http://code.jquery.com/ui/1.10.4/themes/smoothness/jquery-ui.css">
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-49796218-2"></script>
     <script>
-      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-      (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-      m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-      })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
-      ga('create', 'UA-49796218-2', 'mozilla.org');
-      ga('send', 'pageview');
+      var _dntStatus = navigator.doNotTrack || navigator.msDoNotTrack;
```
georgf: Does our GA setup not do DNT automatically?

georgf: If we need our own DNT check, then it should go into a separate function for readability

georgf:
We'll need a follow-up issue to this for outbound link tracking.

georgf:
Our colleague Schalk [blogged about DNT & GA](http://schalkneethling.github.io/blog/2015/11/06/respect-user-choice-do-not-track/), he might be able to give feedback on that if needed.

addon-perf/index.html
```diff
@@ -12,13 +12,26 @@
...
+      } else {
+        _dntStatus = {'0': 'Disabled', '1': 'Enabled'}[_dntStatus] || 'Unspecified';
+      }
+      if (_dntStatus !== 'Enabled') {
+        window.dataLayer = window.dataLayer || [];
+        function gtag(){dataLayer.push(arguments);}
+        gtag('js', new Date());
+
+        gtag('config', 'UA-49796218-2');
+      }
```
wlach: As I said in my review of the previous pr, I think we should factor this function into a seperate file and import it where needed for clarity.

wlach (commented):
#353 had some tracking of  clicking outbound links, which seems like it would be useful here. Aside from that, my only concern is the amount of duplicated code.

chutten:
The DNT code is pulled directly from our data practices mana: https://mana.mozilla.org/wiki/display/DATAPRACTICES/Additional+Cross-Organization+Practices

chutten:
@wlach Ready for another look

analytics/analytics.js
```diff
@@ -0,0 +1,29 @@
...
+    _dntStatus = 'Unspecified';
+  } else {
+    _dntStatus = {'0': 'Disabled', '1': 'Enabled'}[_dntStatus] || 'Unspecified';
+  }
+
+  window.trackingEnabled = () => {
+    return _dntStatus != 'Enabled';
+  }
+  return window.trackingEnabled();
+};
```
wlach, line 20: This construction is a bit hard to understand. Since we're doing this all inside an anonymous function anyway, can you not just unroll this? I also don't see the point in storing trackingEnabled on the window object if we're only going to use it here.

wlach (approved):
I think I'd prefer if some things were a bit simpler, but I'll pre-emptively approve.

chutten:
For the record, the reason I'm hoisting it onto `window` is to enable follow-up patches that will decorate outbound links and do page-specific analytics to know whether or not they should bother. You're entirely correct that there's no reason to do so here :)

PR #353, opened 2017-12-06
Add Google Analytics to TMO

georgf:
To understand what pages are used and by how many users, we are adding GA to the TMO pages.
This also add tracking of outbound links, so we can understand what the landing page is used for.

### DIFF
--- a/dashboard-generator/index.html
+++ b/dashboard-generator/index.html
@@ -13,6 +13,19 @@
 
 <script src="https://telemetry.mozilla.org/v2/telemetry.js"></script>
 <script src="generator.js"></script>
+
+<!-- Global site tag (gtag.js) - Google Analytics -->
+<script async src="https://www.googletagmanager.com/gtag/js?id=UA-110118568-1"></script>
+<script>
+  window.dataLayer = window.dataLayer || [];
+  function gtag(){
+    dataLayer.push(arguments);
+  }
+  gtag('js', new Date());
+
+  gtag('config', 'UA-110118568-1');
+</script>
+
 </head>
 <body>
 <header>
@@ -21,6 +34,7 @@
   <p>This will generate a codepen populated with your dashboard for instant preview and customization. Be sure to click 'Save' on it to show it to others or export for self-hosting.</p>
   <p>For an Introduction, check the <a href="https://chuttenblog.wordpress.com/to-order-telemetry-dashboards-dashboard-generator/">Dashboard Generator Introduction Blog Post</a></p>
 </header>
+
 <main>
   <fieldset>
     <label>Channel:
--- a/dashboard-generator/index.html
+++ b/dashboard-generator/index.html
@@ -15,7 +15,7 @@
 <script src="generator.js"></script>
 
 <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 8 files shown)


georgf:
This worked testing it on my own GA account, i get data flowing in that looks sensible.
I haven't used GA before nor am i experienced in web dev, so i'm not sure if this is a proper approach here.

I did not succeed testing this with the Mozilla telemetry.mozilla.org UA id locally.
The existing TMO tracking for [`addon-perf/`](https://telemetry.mozilla.org/addon-perf/) works with that id already though, so i assume this is because i failed to setup GA properly locally.

Tried to override the domain, but that didn't help:
```gtag('config', 'GA_TRACKING_ID', {
  'cookie_domain': 'mozilla.org', // or telemetry.mozilla.org
});```

**Comment by georgf** [comment]
Hm, so #352 already started on this.
I'm happy to pass this work on, just missed some coordination there.

File: index.html [hunk 1]
```diff
@@ -12,12 +12,44 @@
...
+      var trackOutboundLink = function(obj) {
+        var url = obj.getAttribute("href");
+        console.log('trackOutboundLink', url);
+
+        gtag('event', 'click', {
+          'event_category': 'outbound',
+          'event_label': url,
+          'transport_type': 'beacon',
+          'event_callback': function() {
+            console.log('hitCallback', url);
```
wlach, line 35: Ditto.

index.html
```diff
@@ -12,12 +12,44 @@
...
+      window.dataLayer = window.dataLayer || [];
+      function gtag() {
+        dataLayer.push(arguments);
+      }
+      gtag('js', new Date());
+      gtag('config', 'UA-49796218-2');
+
+      var trackOutboundLink = function(obj) {
+        var url = obj.getAttribute("href");
+        console.log('trackOutboundLink', url);
```
wlach, line 28: Probably want to take this out.

index.html
```diff
@@ -29,45 +61,56 @@ <h2>Telemetry portal</h2>
             <div class="col-md-4">
                 <h3>Main tools</h3>
                 <div class="dashboard-list list-group" style="margin: 0">
-                    <a class="list-group-item" href="new-pipeline/dist.html">
+                    <a class="list-group-item" href="new-pipeline/dist.html"
+                       onclick="trackOutboundLink(this); return false;">
```
wlach, line 65: Could you shorten this to just `return trackOutboundLink(this);` (same for all subsequent)

dashboard-generator/index.html
```diff
@@ -13,6 +13,19 @@
...
+<script async src="https://www.googletagmanager.com/gtag/js?id=UA-49796218-2"></script>
+<script>
+  window.dataLayer = window.dataLayer || [];
+  function gtag(){
+    dataLayer.push(arguments);
+  }
+  gtag('js', new Date());
+
+  gtag('config', 'UA-49796218-2');
+</script>
```
wlach, line 27: Do you think we could consolidate this snippit into a js file that we import from globally? It seems like we repeat this many times.

wlach (commented):
Not really a lot to say here aside from the usual minor coding nits. :) This seems pretty similar to the GA code that I've written myself.

wlach:
Argh, didn't notice that this was closed in favor of another implementation...

PR #357, opened 2017-12-07
Are we supposed to be using gtag instead of ga?

chutten:
The code as committed doesn't seem to send outbound link clicks. Apparently gtag and ga are two different things? Should we be using gtag() instead, then?


wlach:
@chutten I don't actually know :) it's been some time since I've used GA, so am not up to date on the API. Please go ahead and merge if it's helpful to you.

wlach (approved):

Issue #274, opened 2017-01-04
Show statistics with percentiles and median while comparing


digitarald:
The interface has a summary with percentiles and median when showing a single histogram. These numbers would be useful when comparing histograms, but right now the space is left empty.

bani:
I decided to take a stab at this one as an attempt to contribute to this project.

I started with just the Median, as it may become too cluttered to add all percentiles when there is a large number of segments.

You can take a look at [this example on my fork](http://baniverso.com/telemetry-dashboard/new-pipeline/dist.html#!compare=os&cumulative=0&end_date=2017-12-19&keys=__none__!__none__!__none__&max_channel_version=nightly%252F59&measure=GC_MS&min_channel_version=null&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=2017-11-13&table=0&trim=1&use_submission_date=0). Is this what you had in mind?

I'm leaving the order as is (apparently it's alphabetical), but maybe it'd be better to sort by the number. What do you think, is it worth the extra computation?

chutten:
That's pretty good, @bani!

I'm not sure if sorting would be better or worse. All else fails, leave it as it is now and take the sorting question to a follow-up issue.

Now it's just a matter of trying to figure out how to fit the rest of the aggregates in that space...

What do you think about a tabular view with five columns (5th, 25th, median, 75th, 95th) and one row per compare value? It might be one way to fit all the information into the space we have, especially when [comparing many values](https://bit.ly/2DnyhCj).

bani:
Updated to a table.
I had to decrease the font size to fit in the div. Otherwise it was sort of ok, but with a smaller window the scrollbar would cover the 95th percentile column.

chutten:
(Sorry for the delayed response. Just returned from end-of-year holidays)

Smaller window widths are always problematic in the current design. It doesn't relayout properly on narrow screens until it already cuts off quite a bit of content... and that's the current state :S

I wouldn't worry about it too much. We can tackle the problem of narrow-ish windows another time. (You can file an Issue and start working on it right away, if you'd like!)

I'm happy to review it as it is, if you prefer the smaller font size. Or you can return it to the original, your choice.

PR #360, opened 2018-01-04
Histogram comparison statistics

bani:
Resolves https://github.com/mozilla/telemetry-dashboard/issues/274

Display table with percentiles in the empty space next to the histograms when performing a comparison.

@chutten for review

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -4,9 +4,9 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    
+
     <title>Measurement Dashboard</title>
-    
+
     <!-- Bootstrap and supporting libraries -->
     <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css">
     <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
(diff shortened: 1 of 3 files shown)


new-pipeline/dist.html
```diff
@@ -118,6 +118,25 @@ <h3><select id="selected-key4" class="multiselect" title="Selected key"></select
                                 </dl>
                             </div>
                         </div>
+                        <div id="stats-comparison" style="margin-top: 50px; margin-bottom: 20px;">
+                            <table>
+                              <thead class="measure-headers">
+                              <tr>
+                              <th></th>
+                              <th title="5th percentile duration">5 %</th>
```
chutten, line 126: Is there a reason for leaving whitespace between the numbers and the '%'? I wouldn't want the unit to appear on a different line if the table layout algorithm has a bad day :S

bani: I'm a copy&paste HTML developer 😉 
In this case, it was copied from [this addon-perf table](https://github.com/mozilla/telemetry-dashboard/blob/7487b010f29252ce9f260f8250c16dd32e107b8d/addon-perf/index.html#L99)

I imagine it has a space because it's an abbreviation for **percentile**, not a unit.

chutten: I always thought of % as a unit (like Hz, but instead of being "per second" it is "per hundred"). I will have to revise my understanding.

chutten (approved):

chutten:
Thank you for your contribution! It should be live in the next ten minutes or so.

Would you like some help finding the next thing you'd like to work on?

new-pipeline/dist.html
bani: I didn't mean in the general case, it's 100% a unit if considered in the common scenario. :) But then it reads as just percent.
Just in this context it looks like it was a convenient and understandable way to abbreviate percentile when squeezing it in the column title.
So you can keep your understanding unrevised approx. 99% of the time!

PR #369, opened 2018-01-10
Extended the expiry of the 'this is my current dashboard state' cookie to 28 days

### LINKED ISSUE #366: Extend the expiry of the "this is my current dashboard state" cookie
Currently the dashboard state cookie (set in `saveStateToUrlAndCookie()`) is set to expire after only 3 days. 

We are learning from our analytics and user surveys that many users don't visit the dashboard that often.

Let's bump that expiry to a month, maybe?

himanish-star:
fixes : #366 
@chutten , I have made the requested changes. Is 28 days ok for you ?
One thing I would like to ask is that where do I find this line **'this is my current dashboard state'**. I tried searching but couldn't find. And one more question, when does this cookie come into action, as I couldn't login to any of the portals to see what happens there.

### DIFF
--- a/new-pipeline/src/dist.js
+++ b/new-pipeline/src/dist.js
@@ -1728,9 +1728,9 @@ function saveStateToUrlAndCookie() {
       .hide(); // Hide the permalink box again since the URL changed
   }
 
-  // Save the state in a cookie that expires in 3 days
+  // Save the state in a cookie that expires in 28 days
   var expiry = new Date();
-  expiry.setTime(expiry.getTime() + (3 * 24 * 60 * 60 * 1000));
+  expiry.setTime(expiry.getTime() + (28 * 24 * 60 * 60 * 1000));
   document.cookie = "stateFromUrl=" + stateString + "; expires=" + expiry.toGMTString();
 
   // Add link to switch to the evolution dashboard with the same settings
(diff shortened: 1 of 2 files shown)


chutten:
This cookie is for remembering what you were looking at the last time you went to https://telemetry.mozilla.org/new-pipeline/dist.html or https://telemetry.mozilla.org/new-pipeline/evo.html

It will remember which measure, from which channel and version, ... all the things you can change to look at the data it will remember so that you can pick up where you left off last time.

So to test this, go to https://telemetry.mozilla.org/new-pipeline/dist.html or https://telemetry.mozilla.org/new-pipeline/evo.html and change some parameters. Looks at different measures from different channels or OSes. Then start a new tab and go directly to https://telemetry.mozilla.org/new-pipeline/dist.html (or https://telemetry.mozilla.org/new-pipeline/evo.html) and the cookie should restore your state (eventually. It's quite slow).

Obviously we don't want to have you wait for 27 days before testing it again to ensure it works :) It's enough to just make sure that it works within the same session for now.

chutten (commented):
Remove the changes from `src/*.js`. That is a defunct part of the code that we can safely ignore.

himanish-star:
@chutten , Thanks for all this explanation. Now things seem pretty clear and now I can move ahead with more contributions. I have also made the changes requested.

chutten (approved):

chutten:
Please keep the contributions coming! Would you like some help finding something to work on, or did you have something in mind already?

himanish-star:
@chutten , currently I am working on some other issues elsewhere in mozilla . So I'll come back for contributing after maybe a day or two . And thanks for the help. Let me try searching by myself. In case it doesn't work, I'll surely ask you for help.

PR #370, opened 2018-01-15
Getting a lot of "none" urls from analytics.

chutten:
I think we might be getting click events from things that aren't links.

That's unhelpful, so let's filter to register clicks on things that have
hostnames.

### DIFF
--- a/analytics/analytics.js
+++ b/analytics/analytics.js
@@ -36,7 +36,7 @@ function decoratedLinkHandler(e) {
     return;
   }
   let anchor = findAnchor(e.target);
-  if (anchor && anchor.hostname != window.location.hostname) {
+  if (anchor && anchor.hostname && anchor.hostname != window.location.hostname) {
     gtag('event', 'outbound_link', {
       'event_category': 'user_interaction',
       'event_action': 'click',


analytics/analytics.js
```diff
@@ -36,7 +36,7 @@ function decoratedLinkHandler(e) {
     return;
   }
   let anchor = findAnchor(e.target);
-  if (anchor && anchor.hostname != window.location.hostname) {
+  if (anchor && anchor.hostname && anchor.hostname != window.location.hostname) {
```
Dexterp37, line 39: That would also exclude external links, right? Is that intended/expected?

georgf:
I think there is a problem with some real links.

Looking at the [top events](https://analytics.google.com/analytics/web/#report/content-event-events/a49796218w81627824p84513320/%3Fexplorer-table.plotKeys%3D%5B%5D%26explorer-table.rowStart%3D0%26explorer-table.rowCount%3D500%26_r.drilldown%3Danalytics.eventCategory%3Auser_interaction%2Canalytics.eventAction%3Aclick%26explorer-graphOptions.selected%3Danalytics.nthMonth/), i don't see dist.html & evo.html.

Trying things manually and checking the real-time events view:
- I can trigger events for clicking on the links for e.g. hardware report and mission control.
- I can't get events triggered for the distribution or evolution link.

analytics/analytics.js
chutten: It only includes external links. How do you mean they will be excluded?

Dexterp37: Right, sorry for the noise!

chutten:
You won't see dist.html or evo.html in the events. This is specifically for outbound links. Links within TMO are covered by pageviews.

Hardware Report and Mission Control are generating events (they're in top events). Do you perhaps have DNT set? That opts you out from any event generation at all.

georgf:
Chris found that there are `<a>` elements without `href` on dist/evo, specifically the probe selector entries in the dropdown.

chutten:
We don't need to pull this in as we're going to change the methodology in #372

PR #373, opened 2018-01-16
Adjust analytics to collect all links on index, no links elsewhere.

### LINKED ISSUE #372: Adjust Analytics
The analytics as they work now work for answering the questions "What parts of TMO are used?" and "How often?"

It's not so useful for understanding what parts of `TMO/index.html` are most clicked. 

So, we need a change in our analytics. We still need pageviews from across TMO, but now we want to track usage of TMO/index separately. So external links are out. Time to track clicks on all the anchors on the index, including local ones.

Unfortunately, I already used the best name for these events (`click`) so this methodology change will have to use a less-good name. Like `nav`, I guess.

chutten:
Fixes #372

Essentially this just redefines "outbound" links as any anchor that has
a `href` (not just the ones that lead to external sites) and limits our
interest in them to just the index/portal/TMO front.

### DIFF
--- a/analytics/analytics.js
+++ b/analytics/analytics.js
@@ -25,10 +25,17 @@ if (window.trackingEnabled()) {
   gtag('js', new Date());
   gtag('config', 'UA-49796218-2');
 
-  window.addEventListener("DOMContentLoaded", () => {
-    // Decorate outbound links from the top frame
-    document.addEventListener("click", decoratedLinkHandler, true);
-  });
+  // Decorate outbound links from the top frame of whitelisted pages
+  const OUTBOUND_LINKS_WHITELIST = new Set([
+    '/',
+    '/index.html',
+  ]);
+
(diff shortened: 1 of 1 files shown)


analytics/analytics.js
```diff
@@ -25,18 +25,25 @@ if (window.trackingEnabled()) {
...
+  }
 }
 
 function decoratedLinkHandler(e) {
   if (e.button != 0) {
     return;
   }
   let anchor = findAnchor(e.target);
-  if (anchor && anchor.hostname != window.location.hostname) {
+  if (anchor && anchor.href) {
```
georgf: Can we just leave the `href` check out?
Worst-case we just get some more "(not set)" events.
On the plus side, we see when something goes wrong with the instrumentation (as we just did).

georgf (approved):
Looks good to me, with the below addressed.

PR #375, opened 2018-01-18
Ability to download entirety of keyed histogram

### LINKED ISSUE #259: Add ability to download entirety of keyed histogram
Right now, you can only download the keys being displayed.

himanish-star:
fixes : #259 

changes can be previewed here : https://himanish-star.github.io/telemetry-dashboard/

@chutten , I have attached the feature to download the histogram for all the keys. I have made two new buttons : `Export CSV Entire` and `Export JSON Entire`. But one problem I face is that these two buttons are not hidden in case of `non-keyed histograms`. I have used 
``` javascript 
$("#export-csv-entire").hide()
```
but it doesn't work

### DIFF
--- a/new-pipeline/src/dist.js
+++ b/new-pipeline/src/dist.js
@@ -1797,8 +1797,6 @@ function saveStateToUrlAndCookie() {
       // JSON
       var outputObj = {};
       gCurrentHistogramsList
-        .filter(titleHistogramsPair =>
-          shownKeys.indexOf(titleHistogramsPair.title) != -1)
         .sort((a, b) => shownKeys.indexOf(a.title) - shownKeys.indexOf(b.title))
         .forEach(titleHistogramsPair => {
           outputObj[titleHistogramsPair.title] = [];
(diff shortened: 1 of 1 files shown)


new-pipeline/src/dist.js
```diff
@@ -1791,6 +1801,8 @@ function saveStateToUrlAndCookie() {
           }
         )
         .join("\n");
+      $("#export-csv-entire, #export-json-entire")
```
himanish-star:
```javascript
$("#export-csv-entire, #export-json-entire")
 .hide()
```
This code fragment doesn't hide the buttons while viewing a non-keyed histogram

chutten (changes requested):
Instead of adding new buttons, could you augment the existing buttons to download all keys? Downloading only 4 keys is a behaviour we can replace.

himanish-star:
@chutten , Yup doing that now
EDIT : Done, changes can be previewed at https://himanish-star.github.io/telemetry-dashboard

chutten (approved):
Always happy to r+ code removal :)

PR #379, opened 2018-01-18
Documentation for bug 1414839

chutten:


index.html
```diff
@@ -72,6 +72,9 @@ <h3>Documentation</h3>
                     <a class="list-group-item" href="https://wiki.mozilla.org/Firefox/Shield">
                         Shield Docs <div class="dashboard-description">Learn how to run Shield studies &amp; experiments.</div>
                     </a>
+                    <a class="list-group-item" href="doc/ssl.html">
+                        Public SSL Dataset <div class="dashboard-description">Learn what the SSL dataset is and how to use it</div>
```
georgf, line 76: Hm, this is a start for now, but linking to individual dataset docs doesn't really scale.
Is that something we can link off from our data docs instead? Or could we put the whole documentation there?

chutten: Considering the number of public datasets we've released in the past two years, I would argue that this does, unfortunately, scale. :S

Oh wait, now that you mention it, docs.tmo has a section specifically for Datasets. Let me take my documentation over there.

chutten:
Gonna take this to docs.tmo

PR #381, opened 2018-01-20
Data Will Not Be Loaded Until User Is Done Picking Options

### LINKED ISSUE #255: Don't load data until done picking options
The evolutions/histograms UI is too aggressive in updating things before I'm done picking the things I want

himanish-star:
fixes: #255
changes can be previewed at : https://himanish-star.github.io/telemetry-dashboard/

@chutten , I have added a delay of 0.8 s for all options except the `#measure` option in which there is no delay, as I believe all other options are only changed on the basis of the value of the `#measure` option. However I still don't think this is optimal as after 0.8 s the options will get updated and I don't think so it would actually solve the issue. What I intend on doing is to add an event listener for the `select` tags , and thus whenever a user is choosing options from the `select` options data shouldn't be loaded. What do you think? Although I think there are no event listeners to check for whether a user is still scrolling through the options as the only event listener for select is `change`.

### DIFF
--- a/new-pipeline/src/dist.js
+++ b/new-pipeline/src/dist.js
@@ -9,11 +9,27 @@ var gPreviousFilterAllSelected = {};
 var gAxesList;
 var gAxesSelectors;
 var gNoneKey = "__none__";
+var gChangePending = false;
+const kDefaultFilterChangeTimeoutDelay = 3000;
 
 indicate("Initializing Telemetry...");
 
 $(function () {
   Telemetry.init(function () {
+    let btnGroup = $('.btn-group');
+    btnGroup.on('hide.bs.dropdown', function () {
+      if(gChangePending) {
+        setFilterChangeTimeout(function () {
+          $('#measure')
+            .trigger('change');
+        }, kDefaultFilterChangeTimeoutDelay);
+      }
+    });
+    btnGroup.on('click', function () {
+      if(gChangePending) {
+        clearTimeout(gFilterChangeTimeout);
+      }
+    });
     gFilters = {
       "application": $("#filter-product"),
       "os": $("#filter-os"),
@@ -143,10 +159,12 @@ $(function () {
 
       $("#channel-version")
         .change(function () {
-          updateOptions(function () {
-            $("#measure")
-              .trigger("change");
-          });
+          setFilterChangeTimeout(function () {
+            updateOptions(function () {
+              $("#measure")
+                .trigger("change");
+            });
+          }, kDefaultFilterChangeTimeoutDelay);
         });
       $([
         "input[name=build-time-toggle]",
@@ -164,7 +182,7 @@ $(function () {
           if (gFilterChangeTimeout !== null) {
             clearTimeout(gFilterChangeTimeout);
           }
-          gFilterChangeTimeout = setTimeout(function () { // Debounce the changes to prevent rapid filter changes from causing too many updates
+          setFilterChangeTimeout(function () { // Debounce the changes to prevent rapid filter changes from causing too many updates
             if (["filter-product", "filter-os"].indexOf($this
                 .attr("id")) >= 0) { // Only apply the select all change to the product and OS selector
               // If options (but not all options) were deselected when previously all options were selected, invert selection to include only those deselected
@@ -346,7 +364,7 @@ $(function () {
                   .trigger("change");
               }, $("input[name=sanitize-toggle]:checked")
               .val() !== "0");
-          }, 0);
+          }, e.target.id === 'measure' ? 0 : kDefaultFilterChangeTimeoutDelay);
         });
 
       $(
@@ -1884,4 +1902,12 @@ function saveStateToUrlAndCookie() {
 
 function isHistogramsListKeyed(histogramsList) {
   return histogramsList.some(entry => entry.title !== "");
+}
+
+function setFilterChangeTimeout(timeoutFunction, timeoutDelay) {
+  gChangePending = true;
+  gFilterChangeTimeout = setTimeout(() => {
+    timeoutFunction();
+    gChangePending = false;
+  }, timeoutDelay);
 }


himanish-star:
@chutten , could you review my PR please.

chutten:
Yeah, I think I see what you mean. The 800ms delay on its own isn't enough, and needs something clearing `gFilterChangeTimeout` when the next multiselect is opened.

Or, put another way, when `gFilterChangeTimeout` fires, you could check to see if any of the multiselects are open. (and, if one is open, don't trigger the change event)

What do you think?

himanish-star:
Yep, this was what I was trying to convey. I had tried once by adding an event listener to the multiselect but later I realized that only the `change` event listener works for this  ,which fires only when an option is selected from the multiselect menu. So let me retry on trying to implement this

himanish-star:
@chutten , I have something to discuss upon regarding this issue. In my discussion below I have used some `functions` whose description are given below : 

`onDropdownShow`: https://github.com/davidstutz/bootstrap-multiselect/blob/master/dist/js/bootstrap-multiselect.js#L345 (fired when multiselect menu opens)

`onDropdownHide`: https://github.com/davidstutz/bootstrap-multiselect/blob/master/dist/js/bootstrap-multiselect.js#L353 (fired when multiselect menu closes)

 - As we had our earlier discussion it was decided that `gFilterChangeTimeout` will be cleared whenever a multiselect menu is open. Now consider the case when the user first selects an option from a multiselect menu and then opens another multiselect menu before 0.8s lapses and thus this would clear `gFilterChangeTimeout`, then the user actually chooses nothing from this multiselect menu and thus the histogram is not updated as it only updates when an option from any of the multiselect menu is chosen. 
-  So to tackle this I had thought of using the function `onDropdownHide` to make it initiate this(below) which would perform an entire histogram update
```javascript
$('#measure')
    .trigger('change')
```
- The above code works but has a drawback, because  if a user only opens and closes a multiselect menu without choosing anything from the menu then due to this code the entire histogram gets updated

So, could you help me out here. In the meanwhile I'm also thinking of an optimal solution, however some help will surely help me

chutten:
I like the idea of using `onDropdownHide`. The trick there is ensuring that there is a change pending before triggering the `change` event. Otherwise we'll be doing a lot of work every time someone opens and closes a dropdown, regardless of whether anything changed.

So when we clear the timeout, we need to still save some state letting us know that we should restart the timeout `onDropdownHide`.

Make sense?

himanish-star:
Thanks. I'll update you in some time after I make the change.

himanish-star:
I have made the changes now only for `dist.js`. If you find these changes correct then I'll do the same for `evo.js`
You can preview the changes at https://himanish-star.github.io/telemetry-dashboard

new-pipeline/src/dist.js
```diff
@@ -9,9 +9,19 @@ var gPreviousFilterAllSelected = {};
 var gAxesList;
 var gAxesSelectors;
 var gNoneKey = "__none__";
+var changePending = false;
 
 indicate("Initializing Telemetry...");
-
+function stopProgress() {
```
chutten: I'm not a big fan of using globals to reach from JS file to JS file. It's loose coupling that is very fragile.

I wonder, though, if there -is- any way to do this in one place. We may be restrained from that because of the way that `dashboards.js` "helpfully" initializes all the multiselects for us... and the way the multiselect library works.

If we can't do this cleanly, I'd prefer we not do it at all (as helpful as this change would be) :S

Maybe we'll have to rewrite some of the UI before this becomes easily possible. Can you think of a different way to do this that doesn't require skyhooking in some globals from dist.js back into dashboards.js? The dependency on those files generally points the other way (dist.js can depend on things present in dashboards.js, not the other way)

himanish-star: > Can you think of a different way to do this that doesn't require skyhooking in some globals from dist.js back into dashboards.js? 

I'll try my best to think of something else.

chutten: Thank you for your understanding. I'm sorry for being particular about this.

himanish-star: No, problem. I can see why my changes are not optimal. The major trick of this change will be to trigger a function on closing the multiselect-menu. I think I can follow the code of the bootstrap-multiselect to see how this is triggered and that may be the key to my solution

himanish-star: @chutten , Have a look now

new-pipeline/src/dist.js
```diff
@@ -9,11 +9,30 @@ var gPreviousFilterAllSelected = {};
 var gAxesList;
 var gAxesSelectors;
 var gNoneKey = "__none__";
+var changePending = false;
```
chutten: Style in this file is to prepend with `g` any global state. (so, `gChangePending`)

new-pipeline/src/dist.js
```diff
@@ -9,11 +9,30 @@ var gPreviousFilterAllSelected = {};
...
+  }
+}
 $(function () {
   Telemetry.init(function () {
+    let btnGroup = $('.btn-group');
+    btnGroup.on('hide.bs.dropdown', function () {
+      startProgress();
+    });
+    btnGroup.on('click', function () {
+      stopProgress();
```
chutten: We can inline `stopProgress` in here, to avoid creating another top-level function the interpreter has to parse at init.

new-pipeline/src/dist.js
```diff
@@ -9,11 +9,30 @@ var gPreviousFilterAllSelected = {};
...
+        .trigger('change');
+      changePending = false;
+      }, 3000);
+  }
+}
 $(function () {
   Telemetry.init(function () {
+    let btnGroup = $('.btn-group');
+    btnGroup.on('hide.bs.dropdown', function () {
+      startProgress();
```
chutten: We can inline `startProgress` here, since this is the only place it's called.

new-pipeline/src/dist.js
```diff
@@ -143,10 +162,14 @@ $(function () {
 
       $("#channel-version")
         .change(function () {
-          updateOptions(function () {
-            $("#measure")
-              .trigger("change");
-          });
+          changePending = true;
+          gFilterChangeTimeout = setTimeout(function () {
```
chutten: We could factor out this common pattern and reuse it in the three places we need it:

```js
function setFilterChangeTimeout(timeoutFunction, timeoutDelay) {
  gChangePending = true;
  gFilterChangeTimeout = setTimeout(() => {
    timeoutFunction();
    changePending = false;
  }, timeoutDelay);
}
```

(the place to put it would be at the bottom)

new-pipeline/src/dist.js
```diff
@@ -9,11 +9,30 @@ var gPreviousFilterAllSelected = {};
...
+      }, 3000);
+  }
+}
 $(function () {
   Telemetry.init(function () {
+    let btnGroup = $('.btn-group');
+    btnGroup.on('hide.bs.dropdown', function () {
+      startProgress();
+    });
+    btnGroup.on('click', function () {
```
chutten: Is there a `'show.bs.dropdown'` we could use, for symmetry's sake?

new-pipeline/src/dist.js
```diff
@@ -9,11 +9,30 @@ var gPreviousFilterAllSelected = {};
...
+function stopProgress() {
+  clearTimeout(gFilterChangeTimeout);
+}
+function startProgress() {
+  if(changePending) {
+    gFilterChangeTimeout = setTimeout(function () {
+      $('#measure')
+        .trigger('change');
+      changePending = false;
+      }, 3000);
```
chutten: We should make this a constant. `const kDefaultFilterChangeTimeoutDelay = 3000;` That way it'll be easier to tune if we find that 3s ends up being too long or too short.

chutten (commented):
I like it! I have a few comments about the implementation, but the idea seems very sound!

new-pipeline/src/dist.js
himanish-star: Yes, there is a `show.bs.dropdown`, but I can't use that here. 
_____
https://github.com/mozilla/telemetry-dashboard/blob/66ed4e9dcd263f563bd8361fba8a507e2c19367c/lib/bootstrap-multiselect.js#L353
Here on `show.bs.dropdown`(triggered by metricsgraphics.js) the function `onDropdownShow` is triggered which is used here
https://github.com/mozilla/telemetry-dashboard/blob/f66351757f13829807efd3c661ae28a0bcf5899a/new-pipeline/src/dashboards.js#L18
So if I add an event listener for the same in `dist.js` then this would overwrite this `onDropdownShow` function
______
https://github.com/mozilla/telemetry-dashboard/blob/66ed4e9dcd263f563bd8361fba8a507e2c19367c/lib/bootstrap-multiselect.js#L354
But here no overriding takes place as the `onDropdownHide` function is never used, thus I have used it

new-pipeline/src/dist.js
himanish-star: Ok, I'll do that

new-pipeline/src/dist.js
himanish-star: Sure, I'll do this too

new-pipeline/src/dist.js
himanish-star: Ok

himanish-star:
@chutten, I have made the changes please review

chutten (approved):
I like it!

Do you have this up and running somewhere I can get the feel of the 3s delay? It seems really long in the code, but I don't know if it _feels_ long when a user is actually interacting with it.

himanish-star:
Yes, you can preview them at https://himanish-star.github.io/telemetry-dashboard.
But there are some problems
The timeout is not cleared in case of all multiselect menu. I don't know why the timeout is not cleared in this case
However timeout works fine for the single select menu. Timeout is cleared in this case
EDIT: Please help me out on clearing timeout. As the `clearTimeout` does fire but  fails in the case of the multiselect options like architecture, e10s and OS. But passes in options like compare, channel-version and process-type because of them being single select

chutten:
The live example is indeed having problems. I seem to be able to get into a state where even changing the measure doesn't result in new data being requested and the plots redrawn.

himanish-star:
@chutten, have a look now. I hadn't updated the gh-pages branch

chutten:
I see what you mean. If I change from all OS to just All Windows, then open the channel/version selection, then three seconds later it updates to the "all windows" data without waiting for me to make my channel/version selection.

himanish-star:
Yeah but things work for content process and the compare with option. I tried a lot to identify why things work perfect here but not with the other options but couldn't make way

himanish-star:
you can select something from the process option and then select any other option and you will see that things work pretty well

chutten:
I think I may have found a clue.

By logging in the click handler (for the `clearTimeout`) and in `setFilterChangeTimeout` I noticed that interacting with the multiselect triggers `setFilterChangeTimeout` immediately, without waiting for the dropdown to be dismissed.

It triggers a second filter change timeout when the dropdown is dismissed. The click handler can only cancel one of the two timeouts (since the other has been or will be overwritten), so the other timeout still happens 3s later.

himanish-star:
Cool thanks. I knew something like this was happening but couldn't locate it. I'll try fixing this now

chutten:
@himanish-star It's been a little while, are you having problems? Can I help?

himanish-star:
@chutten, actually I was a little busy these days. I will get back to this tomorrow for sure

chutten:
Oh good. Get to it when you can, I just was making sure you weren't toiling away in frustrated silence :)

himanish-star:
:laughing: Actually I had a number of events in college. So that is why I hadn't quite looked into this. I will surely send you updates on this tonight.

himanish-star:
@chutten, Sorry for delaying this PR so much but the thing is that I am actually applying for GSoC this year and thus I'm quite busy writing the proposal for the same. Adding to this I have my college exams coming up and thus I am unable to devote time to this PR. So please pardon me for not updating my work.

chutten:
I understand. We're very happy to have any contribution you have the time for, but if you don't have the time for it, don't worry! School and life come first.

We can leave this open if you think you might have time for this later. And if not, we can keep this around as a starting point for someone else. It won't be wasted :)

himanish-star:
Ok I'll be taking 2 days off from college this week. So I'll try to look into this today if possible or on Saturday for sure

chutten:
I'm going to close this because of inactivity. We can reopen or adapt this solution later.

Sylvia23:
@chutten , Can I take this up to complete it so that it doesn't get wasted.

chutten:
@Sylvia23 Please do!

Issue #350, opened 2017-12-04
Directly link to detail view for data explorer


georgf:
We added linking to the data explorer with #349.
In [this PR](https://github.com/georgf/fx-data-explorer/pull/14) i'm adding a detail view.
We can link to that directly using e.g. `...?detailView=histogram%2FGC_MS`.

georgf:
We can directly link for histograms & scalars, using: `detailView=<type>/<name`.
For a histogram like [GC_MS](https://mzl.la/2mSkwEn) this is `detailView=histogram%2FGC_MS`.
For e.g. a scalar like [this](https://mzl.la/2mSjOqH) this is `detailView=scalar%2Fbrowser.engagement.max_concurrent_tab_count`.

`SIMPLE_MEASURES` / data from `simpleMeasurements` is not covered yet, we should just avoid linking for simple measures and other unsupported things.
Some environment data is there, but not in the dash here as far as i can see.

georgf:
Fixed by #384.

PR #384, opened 2018-01-20
Improve detail links

georgf:
This addresses #350.

- Don’t add a “more details” link for SIMPLE_MEASURES_*, as the
probe-dictionary doesn’t have them.
- Directly link to the detailView for everything else (should be all
scalars & histograms).

At some point we’ll have to know the actual probe type (histogram,
scalar, …), but this seem to work for now.


new-pipeline/src/dashboards.js
```diff
@@ -1046,16 +1046,52 @@ function updateOSs() {
...
+    descr = description;
+  }
 
   var div = document.createElement('div');
   div.classList.add("text-center");
-  div.innerHTML =  descr + " <a href=\"" + metricUrl + "\" target=\"_blank\">" +
-                   "<i class=\"btn btn-outline-primary fa fa-info-circle\" " +
-                   "style='color: black' aria-hidden=\"true\">  " +
-                   "More Details </i></a>"
+  var html = descr;
```
chutten: Using `innerHTML`, you should wrap this in a tag. Or we should build the nodes using DOM APIs since metricURL contains user-supplied input, then we can avoid `innerHTML` entirely.

chutten (approved):
One smallish thing

georgf:
@chutten Let's give this another round of review.

chutten (approved):

Issue #376, opened 2018-01-18
wrapper: categorical histogram labels are "NaN"


leplatrem:
![screenshot from 2018-01-18 10-59-11](https://user-images.githubusercontent.com/546692/35092104-a866b774-fc3e-11e7-9873-b833418c5695.png)

> Note: This is a keyed categorical histogram

See https://codepen.io/anon/pen/LeXOyb

leplatrem:
Some `else if categorical` missing here ?

https://github.com/mozilla/telemetry-dashboard/blob/aab8725f543581b90b41e89a8fa589b807fe82e3/wrapper/telemetry-wrapper.js#L512-L513

chutten:
Almost certainly. Wrapper pre-dates categorical histograms.

PR #386, opened 2018-01-21
Render categorical histograms correctly.

o0Ignition0o:
Trying to figure out a -not too ugly- way to display the correct keys for #376. 

screenshot from 2018-01-21 18-02-53

The keys seem to be correct, although it looks really bad ^^'. 
I don't really know what would be the most convenient layout (Maybe using acronyms instead ? A vertical display?) I have literally no taste so I would be more than happy to take any suggestions :)

I've added some really simple if statements in order to be able do show you a screenshot asap, but as stated by the telemetry documentation (https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/histograms.html?highlight=categorical) there seem to be lots of allowed histogram kinds : 
``` 
boolean
linear
exponential
categorical
enumerated
flag
<deprecated> count 
``` 
I'm considering using switch / case statements to prepare the bucket labels (and maybe DRYing the code a little bit more), but I don't know the codebase enough to be able to judge if it would be useful / worth it.
@chutten @leplatrem What do you think ?
I would love to dig deeper on this if someone can mentor me :)

EDIT: After searching a bit more it seems like categorical is the only entry which wouldn't require any formatNumber() call, so I guess ifs are fine. 
As displayed here : https://telemetry.mozilla.org/histogram-simulator/#low=1&high=500&n_buckets=20&kind=exponential&generate=normal a vertical display seems the most used approach, so I'll update the PR and stay open for any suggestions :)

### DIFF
--- a/wrapper/telemetry-wrapper.js
+++ b/wrapper/telemetry-wrapper.js
@@ -496,6 +496,9 @@ window.TelemetryWrapper.go = function (params, element) {
         if (!starts[index]) {
           return '';
         }
+        if (hist.kind == 'categorical') {
+          return starts[index];
+        }
         return formatNumber(starts[index]);
       },
       yax_format: value => value + '%',
@@ -511,6 +514,8 @@ window.TelemetryWrapper.go = function (params, element) {
           label = ' \u2265 ' + formatNumber(starts[index]);
         } else if (hist.kind == 'enumerated') {
           label = formatNumber(starts[index]);
+        } else if (hist.kind == 'categorical') {
+          label = starts[index];
         } else {
           label = '[' + formatNumber(starts[index])
                   + ', ' + formatNumber(ends[index]) + ')';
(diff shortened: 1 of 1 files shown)


o0Ignition0o:
screenshot from 2018-01-21 18-50-33
FYI a 90° rotation on the x_axis labels would require growing the diagrams height, which could have quite an impact for anyone embedding histograms. 

I think few users are embedding categorical histograms yet, and resizing the graph is less of a pain than displaying NaN labels though.

chutten:
On https://mzl.la/2DpPWsI we angle them by 20deg using some [SVG hackery](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/new-pipeline/src/dist.js#L1543).

Emulating this look would make it more consistent.

leplatrem:
> On https://mzl.la/2DpPWsI we angle them by 20deg using some SVG hackery.

Indeed that's a lot easier to read than 90° :) 

Thanks for tackling this!

o0Ignition0o:
screenshot from 2018-01-24 18-53-15
Ok I'm almost there, I just need to figure out a way to move the graph titles and adjust the svg height :)

o0Ignition0o:
screenshot from 2018-01-24 21-06-36
The numbers I've chosen are totally arbitrary (and magic) so we might want to change them or give them a name.
Also the graphContainer resizing feels a bit "hacky" because I don't feel like I'm supposed to do it there, but I haven't figured out a better way to handle this, so I'm open for suggestions :)

Else I guess it can be r? @chutten :)

wrapper/telemetry-wrapper.js
```diff
@@ -542,6 +547,28 @@ window.TelemetryWrapper.go = function (params, element) {
     for (var tick of graphEl.querySelectorAll('.mg-extended-y-ticks')) {
       tick.setAttribute('x2', window.parseFloat(tick.getAttribute('x2')) + missingWidth);
     }
+
+    if (hist.kind == 'categorical') {
+      // Rotate the x axis labels
```
chutten: Please introduce a comment up here that the following code contains magic values that seem to work for now.

chutten (commented):
Could you edit your commit message to be clear that this is just for categorical histograms rendered by telemetry-wrapper? [wrapper] at the beginning would be enough for me.

o0Ignition0o:
Is it now explicit enough ? :)

chutten (approved):
Very good, thanks!

PR #391, opened 2018-01-24
metricsgraphics.min.js recommended instead of metricsgraphics.js

### LINKED ISSUE #390: Update telemetry-wrapper's documentation to recommend using the minified MG
telemetry-wrapper's docs recommend the use of `metricsgraphics.js` when we also have a much smaller and faster-to-parse `metricsgraphics.min.js`

We should update the docs to recommend it instead.

himanish-star:
fixes: #390


chutten:
Thank you for your contribution!

PR #393, opened 2018-01-30
short-description : Fixes #383

### LINKED ISSUE #383: Limit the visible length of the probe descriptions
The probe descriptions can have arbitrary length.
We should hide them after a certain length (one line on reasonable window widths?).
We can consider making them expand on user actions.

shikhar-scs:
As mentioned by @georgf at #383, the description is now shortened and the length is adjustable.

Here's a preview

short

Please review.

### DIFF
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -1077,7 +1077,14 @@ function getDescriptionWithLink(metric, channel, description) {
   var metricUrl = buildDictionaryURL(metric, channel, description);
   var descr = metric;
   if (description && (description.length > 0)) {
-    descr = description;
+	  
+	  if (description.length > 200) {
+		  descr = description.substring(0, 200);
+		  descr += ' ...'
+	  }
+	  else {
+		  descr = description;
+	  }
   }
 
   var div = $("<div>", {


new-pipeline/src/dashboards.js
```diff
@@ -1077,7 +1077,14 @@ function getDescriptionWithLink(metric, channel, description) {
   var metricUrl = buildDictionaryURL(metric, channel, description);
   var descr = metric;
   if (description && (description.length > 0)) {
-    descr = description;
+	  
+	  if (description.length > 200) {
+		  descr = description.substring(0, 200);
+		  descr += ' ...'
```
georgf, line 1083: This has roughly the desired effect, but the better approach is to solve this using CSS.
We should probably use styling with `text-overflow: ellipsis;` etc. to achieve this effect.
https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

shikhar-scs: Um, ok :+1:

shikhar-scs:
@georgf 
I tried the other approach. For something like `text-overflow: ellipsis;` to work, as mentioned on the link you provided me above, there needs to exist the following css properties in its parent div 

```
.overflow-div {
  overflow: hidden;
  white-space: nowrap;
}
```

Removing either of those makes the `ellipses` property redundant. But what I achieved from that is this

screenshot from 2018-01-31 08-28-22

The line gets cut abruptly, and `More Details` remains hidden

georgf:
We could change how the elements are structured & styled to fix this, that is fine.
The pieces that affect this are:
- `getDescriptionWithLink()` in dashboards.js
- the element `#dist-caption` in dist.js & dist.html (and surrounding elements)
- the element `#measure-description` in evo.js & evo.html (and surrounding elements)
- the corresponding CSS

georgf:
Hi Shikar, are you stuck with anything? Can i help?

shikhar-scs:
Hey george !!! I've actually been a bit pre occupied with so much work . Thanks for asking though. I'll get back to work in a few hours 😅😅

georgf:
Hi shikhar, would it be ok for someone else to take over this issue for now?
Happy to make it available to you again or to have you active on other bugs!

shikhar-scs:
Yeah sure. Actually GSOC is keeping me a bit busy these days. With completely new projects, I have to invest a lot of time there. I'll be happy to be back once I get over with it. :sweat_smile: Sorry, for all the inconvenience. As of now, yeah anyone can take over this. No issues. :+1:

georgf:
This was solved by #418 :)

PR #402, opened 2018-02-21
Fix #401: Include protocols for referenced libs

### LINKED ISSUE #401: Replace protocol-less URLs with HTTPS URLs
Currently when loading an HTML file from the repo locally using `file://`, the URLs referenced that use a protocol-less format do not work as the resources come from the internet.

I propose that we replace these with full URLs with the `https://` protocol so these work both locally (via `file://` loading) and on TMO (which uses HTTPS).

robhudson:
This broke local viewing of TMO. Also, TMO is always served from https
so we should also link to the secure versions of these resources.

### DIFF
--- a/histogram-simulator/index.html
+++ b/histogram-simulator/index.html
@@ -7,10 +7,10 @@
     
     <title>Histogram Simulator</title>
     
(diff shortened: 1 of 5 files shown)


chutten (approved):
LGTM

Issue #407, opened 2018-02-26
Move landing page column with twitter feed to the right


georgf:
On the [landing page](https://telemetry.mozilla.org/), the middle column contains a twitter feed.
User feedback tells us that a) this is not useful to most and b) that they tend to miss the other tools in the right-most column.

In `/index.html`, we should move the twitter column to the right, after the column that contains "getting help" and "other tools".

Sylvia23:
Hi @georgf , Can you please review the PR. Thanks!

georgf:
Fixed by #410.

PR #410, opened 2018-02-26
fix for #407 : Move landing page column with twitter feed to the right 

Sylvia23:
Hi @georgf , this is the fix for issue 407. Can you please review it? Thanks!

### DIFF
--- a/index.html
+++ b/index.html
@@ -146,7 +146,7 @@
                         data-chrome="noheader nofooter noborders noscrollbar transparent"
                         data-widget-id="410497663147053056">
                         Loading updates on telemetry...
-                    </a>
+                    </a></br>
                     <a href="https://twitter.com/MozTelemetry"
                        class="twitter-follow-button"
                        data-show-count="false"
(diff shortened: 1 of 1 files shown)


o0Ignition0o (commented):
Hi @Sylvia23 and thanks for your submission :)

This looks really good, but there's one tiny nit : as you can see [here](https://github.com/mozilla/telemetry-dashboard/commits/gh-pages) We're trying to write commit messages which allow us to understand the patch reasons as soon as we check the git log. 
Something like `  Fix #407: Move landing page column with twitter feed to the right  ` would be great ! 
I usually do this by using the `git rebase -i` command, which can sometimes be tricky. 
If you need any help please let me know :)

Sylvia23:
Thanks @

*o0Ignition0o,*
Yeah , I would definitely push it with proper message.

On Tue, Feb 27, 2018 at 12:25 AM, Jeremy Lempereur <notifications@github.com

georgf (approved):
This looks good to me, thanks for the contribution!

Issue #411, opened 2018-02-26
Remove e10s setting from telemetry dashboards


georgf:
e10s (using multiple processes) is the standard in Firefox now, so it's not useful anymore to break this out anymore in the dashboard.
We should remove this from the front-end:
![e10s](https://user-images.githubusercontent.com/407696/36686376-84047608-1b26-11e8-8507-cf6ba679fc7c.png)

This potentially affects evo.html, evo.js, dist.html, dist.js & dashboards.js in `/new-pipeline`.

Sylvia23:
Hi @chutten , Can you please review the new PR on this issue. Thanks!

chutten:
I don't see a new PR. It might be best to (force-)push to the old PR, so we keep the context together.

Sylvia23:
Ohh. Okay @chutten 
I will push again.

Sylvia23:
Hi @chutten ,  #413 
 Can you view this PR ?

PR #413, opened 2018-02-26
Fix for #411 : Remove e10s setting from dashboard

Sylvia23:
Hi, Can you please review the PR, and suggest any changes if required. Thanks!

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -47,8 +47,6 @@
                     <select id="filter-os" class="multiselect" title="Target OS" data-all-selected="any OS" data-n-selected="OSs" multiple></select>
                     <select id="filter-arch" class="multiselect" title="Product architecture" data-all-selected="any architecture" data-n-selected="architectures" multiple></select>
                     with
-                    <select id="filter-e10s" class="multiselect" title="e10s setting" data-all-selected="any e10s setting" multiple></select>
-                    as
                     <select id="filter-process-type" class="multiselect" title="Process type" data-all-selected="any process type"></select>
                     and compare by
                     <select id="compare" class="multiselect" title="Filter name to compare options for">
@@ -56,7 +54,6 @@
                         <option value="os">target OS</option>
                         <option value="osVersion">target OS version</option>
                         <option value="architecture">product architecture</option>
-                        <option value="e10sEnabled">e10s setting</option>
                         <option value="child">process type</option>
                     </select>
                 </h2>
--- a/new-pipeline/src/evo.js
+++ b/new-pipeline/src/evo.js
@@ -53,7 +53,6 @@ $(function () {
       "application": $("#filter-product"),
       "os": $("#filter-os"),
       "architecture": $("#filter-arch"),
-      "e10sEnabled": $("#filter-e10s"),
       "child": $("#filter-process-type"),
     };
     gInitialPageState = loadStateFromUrlAndCookie();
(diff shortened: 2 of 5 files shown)


index.html
```diff
@@ -73,27 +73,6 @@ <h3>Documentation</h3>
                     </a>
                 </div>
             </div>
-            <div class="col-md-4">
```
chutten: It seems as though there's some unrelated work here that will need to be omitted from this PR.

new-pipeline/dist.html
```diff
@@ -181,7 +179,7 @@ <h3><select id="selected-key4" class="multiselect" title="Selected key"></select
                                     <label><input name="build-time-toggle" type="radio" value="1"> Submission Date</label>
                                 </div>
                                 <div>
-                                    <b>Sanitize data:</b> hide data points that don't have enough submissions:
+                                    <b>sanitize-toggletize data:</b> hide data points that don't have enough submissions:
```
chutten: I'm not sure this was an intentional change?

new-pipeline/dist.html
```diff
@@ -47,8 +47,6 @@ <h2 style="font-size: 18px; margin-bottom: 0">
                     <select id="filter-os" class="multiselect" title="Target OS" data-all-selected="any OS" data-n-selected="OSs" multiple></select>
                     <select id="filter-arch" class="multiselect" title="Product architecture" data-all-selected="any architecture" data-n-selected="architectures" multiple></select>
                     with
-                    <select id="filter-e10s" class="multiselect" title="e10s setting" data-all-selected="any e10s setting" multiple></select>
-                    as
```
chutten, line 51: After removing this, I would expect that the javascript will start throwing errors when it can no longer find the multiselect. Are there errors in the log?

chutten (changes requested):
Thank you for your contribution! There are some things that will need to be cleared out of here before we can have this merged... and I think there might be some JS changes yet to be made.

Also, the evolution dashboard has an e10s selector that I hope we can get to in this PR as well. Have you seen it?

Sylvia23:
Yes, I have seen the same in evolution dashboard as well. I will remove from the evolution dashboard also. Thanks!

index.html
Sylvia23: Sorry, This was unintentional.

new-pipeline/dist.html
Sylvia23: Sorry, This was unintentional.

Sylvia23:
@chutten , Can you please review this. Thanks!

georgf:
I took this for a quick test locally and the distribution view and evolution view seem to work fine.
I did not look at the code changes.

new-pipeline/src/evo.js
```diff
@@ -1,4 +1,4 @@
-var gInitialPageState = null;
+  var gInitialPageState = null;
```
chutten: Unintentional whitespace change, I think

chutten (commented):
Just one whitespace change and I think we'll be good to go.

new-pipeline/src/evo.js
Sylvia23: Yeah ! Thanks!

Sylvia23:
@chutten , I have removed the whitespace which I put unintentionally. I think we are now good to go.
Can you please review the PR.
Thanks!

chutten (approved):

chutten:
I think we're good to go. I've pushed the change into the tree and, if we've fixed the deploy script properly, we should see this live in production in a few minutes.

PR #415, opened 2018-02-26
Referencing #409 Clearly show the distribution and evolution views in dashboard 

### LINKED ISSUE #409: Show the distribution and evolution views as tabs in dashboard

dashboard mockup

Chauhan99:
Fixes #409


georgf (commented):
Thanks for the contribution Chauhan!

I was trying out the changes locally, but the visualization did not load.
Looking into the [web console](https://developer.mozilla.org/en-US/docs/Tools/Web_Console/Opening_the_Web_Console) i saw the following error:
> uncaught exception: `fromVersion` and `toVersion` must be strings

georgf:
Would you like help to dig into this error or try investigating this on your own first?

georgf:
There hasn't been activity here for a few days, so closing this for now.
Let us know if you want to look at some issues, we can re-open this or point you to other issues!

PR #416, opened 2018-02-26
Fix #409 : Clearly show the distribution and evolution views in dashboard

### LINKED ISSUE #409: Show the distribution and evolution views as tabs in dashboard

Sylvia23:
Error message is deleted from Measurement Dashboard.


georgf:
This looks like this PR includes changes from other PRs and no changes for issue #409 yet.

Issue #406, opened 2018-02-26
Use different styles for measure heading if description is available


georgf:
Comparing [a probe without description](https://mzl.la/2FxhYUN) and [a probe with description](https://mzl.la/2FwPive), you'll notice that one has a heading with the probe name, the other has a description.

What we should have instead is:
- always show the probe name with a "heading" style
- if there is a description, show a text section with it with a normal/regular font style

Shruti9520:
Hi @georgf 
Can I work on this issue? :)

Thanks.

georgf:
Sure, welcome!
Do you want to start digging into the code by yourself or would you like some pointers?

Shruti9520:
I started looking into codebase. Some pointers would be appreciated. :)

georgf:
#393 and #383 touch the same dashboard area, so this will potentially affect dashboards.css, dist.html, evo.html, evo.js, dist.js in `/new-pipeline`.
In Firefox, using right-click and "inspect element" should get started with where this is used in HTML.

Shruti9520:
@georgf Thanks for the help. 
I have a doubt. I am sorry if I sound stupid.
When I run - `python -m http.server`, it gives me an error  "No module named http". So, I tried running ` python -m SimpleHTTPServer 8000`. But this doesn't help. Will you mind telling me the exact problem?

Thanks

Shruti9520:
@georgf I tried running a remote webserver using `YOUR_USERNAME.github.io/telemetry-dashboard` after pushing changes to gh-pages branch. So it looks like this-
![mozilla_pr](https://user-images.githubusercontent.com/20228417/36812219-6adf465e-1cf6-11e8-8744-a52e5be44239.png)

Is that Okay?

georgf:
Personally, for local testing i just:
- change my working directory to be inside the telemetry-dashboard directory
- `python -m SimpleHTTPServer 8000`
- in Firefox, navigate to `localhost:8000` or e.g. `localhost:8000/new-pipeline/dist.html`

Does that work for you? Do you run into any specific errors?

georgf:
The screenshot looks like a great start!
The description text could use a smaller font size, probably just using the regular text size used elsewhere should be fine.

Shruti9520:
@georgf It works fine. Thanks :)
Looking forward to submitting a PR.

georgf:
Great, let us know if you need any further help!

Shruti9520:
[...]

PR #418, opened 2018-02-27
Fix #383 : Limit the visible length of the probe descriptions

### LINKED ISSUE #383: Limit the visible length of the probe descriptions

Sylvia23:
If a probe has long caption, then it doesn't look good if shown completely in the UI. This PR has resolved this issue. Knowledge of HTML, CSS(especially styling `text-overflow: ellipsis;` ) and core JavaScript is required.

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -62,7 +62,9 @@
                 </h2>
                 <hr></hr>
                 <figure class="col-md-12">
-                  <figcaption id="dist-caption"></figcaption>
+                  <figcaption id="dist-caption" class="caption-div"></figcaption>
+                  <figcaption id="caption-dots" class="caption-div"></figcaption>
+                  <figcaption id="dist-caption-link" class="caption-div"></figcaption>
                     <div class="col-md-9" id="plots">
                         <div class="row">
                             <div class="col-md-6" style="text-align: center">
--- a/new-pipeline/evo.html
+++ b/new-pipeline/evo.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+    <!DOCTYPE html>
 <html lang="en">
 <head>
     <meta charset="utf-8">
@@ -60,7 +60,9 @@
                     as
                     <select id="filter-process-type" class="multiselect" title="Process type" data-all-selected="any process type"></select>
                 </h2>
-                <h3 id="measure-description" style="font-size: 14px; margin: 0 3px"></h3>
+                <figcaption id="evo-caption" class="caption-div"></figcaption>
+                <figcaption id="caption-dots" class="caption-div"></figcaption>
+                 <figcaption id="evo-caption-link" class="caption-div"></figcaption> 
                 <h2 style="font-size: 18px; margin-bottom: 0">Showing graphs for <select id="selected-key" class="multiselect" title="Selected key"></select></h3>
                 <figure class="col-md-12">
                     <div class="col-md-12" id="evolutions"></div>
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -1073,8 +1073,7 @@ function buildDictionaryURL(metric, channel, description) {
   return baseUrl + "?" + $.param(params);
 }
 
-function getDescriptionWithLink(metric, channel, description) {
-  var metricUrl = buildDictionaryURL(metric, channel, description);
+function getDescription(metric, channel, description) {
   var descr = metric;
   if (description && (description.length > 0)) {
     descr = description;
@@ -1084,6 +1083,16 @@ function getDescriptionWithLink(metric, channel, description) {
     class: "text-center",
   });
   div.append($("<span>").text(descr));
+
+  return div;
+}
+
+function getDescriptionLink(metric, channel, description) {
+  var metricUrl = buildDictionaryURL(metric, channel, description);
+
+  var div = $("<div>", {
+    class: "text-center",
+  });
 
   if (metricUrl) {
     var link = $("<a>", {
--- a/new-pipeline/src/dist.js
+++ b/new-pipeline/src/dist.js
@@ -886,9 +886,12 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
     var description = histogramsList[0].histograms[0].description
     var metric = histogramsList[0].histograms[0].measure
     var channel = $("#channel-version").val().split("/")[0]
-    var desc = getDescriptionWithLink(metric, channel, description)
+    var desc = getDescription(metric, channel, description)
+    var link = getDescriptionLink(metric, channel, description)
 
     $('#dist-caption').html(desc);
+    $('#caption-dots').html("...");
+    $('#dist-caption-link').html(link);
   } else {
     $('#dist-caption').text(""); // Clear the histogram caption
   }
--- a/new-pipeline/src/evo.js
+++ b/new-pipeline/src/evo.js
@@ -333,11 +333,13 @@ $(function () {
 
               var metric = $("#measure").val()
               var channel = $("#max-channel-version").val().split("/")[0]
-              var description = getDescriptionWithLink(metric, channel, evolutionDescription)
-
+              var description = getDescription(metric, channel, evolutionDescription)
+              var link = getDescriptionLink(metric, channel, description)
               $("#submissions-title").text(metric + " submissions");
               $("#sample-counts-title").text(metric + " sample counts");
-              $("#measure-description").html(description);
+              $("#evo-caption").html(description);
+              $('#caption-dots').html("...");
+              $('#evo-caption-link').html(link);
               $("#selected-key")
                 .trigger("change");
             });
--- a/new-pipeline/style/dashboards.css
+++ b/new-pipeline/style/dashboards.css
@@ -126,10 +126,26 @@
 .mg-marker-text { font-size: 12px !important; }
 .mg-markers line { stroke-width: 2px !important; }
 
-#dist-caption {
+.caption-div {
+  overflow-x: hidden; 
+  display: inline-block;  
+}
+
+#caption-dots {
+  width : 2%;
+  font-size: 1.5em;
+}
+
+#dist-caption, #evo-caption {
+  text-overflow: ellipsis;
+  white-space: nowrap;
   font-size: 1.5em;
   text-align: center;
-  margin: 0.1em;
+}
+
+#dist-caption-link, #evo-caption-link {
+  color : black;
+  font-size: 1.5em;
 }
 
 /* Error message styles */


georgf (commented):
This looks good visually!
I'm mostly wondering whether we need the width limit here, the rest are smaller style issues etc.

Sylvia23:
Hi @georgf , As of now, I have commited this PR without the changes you suggested.
I will change the code as per you suggestions and then commit again.
Thanks!

georgf:
Sounds good, thanks.

Sylvia23:
@georgf , Unfortunately, I couldn't see the changes that you suggested. It seems that they have gone. I can only see the comments in the mail but not the references. So, I am unable to figure out the suggested changes.
Can you please review it again. I am extremely sorry for the inconvenience.
Thanks!

georgf:
No worries, will do!

new-pipeline/evo.html
```diff
@@ -60,7 +60,7 @@ <h2 style="font-size: 18px; margin-bottom: 0">
                     as
                     <select id="filter-process-type" class="multiselect" title="Process type" data-all-selected="any process type"></select>
                 </h2>
-                <h3 id="measure-description" style="font-size: 14px; margin: 0 3px"></h3>
+                <h3 id="measure-description"></h3>
```
georgf: Nice cleanup!

new-pipeline/src/dashboards.js
```diff
@@ -1077,7 +1077,13 @@ function getDescriptionWithLink(metric, channel, description) {
   var metricUrl = buildDictionaryURL(metric, channel, description);
   var descr = metric;
   if (description && (description.length > 0)) {
-    descr = description;
+    if (description.length > 85) {
+      descr = description.substring(0, 85);
```
georgf: Do we need this JavaScript change?
It would be great if we can do this just in CSS, so the dashboard can e.g. automatically show longer descriptions for larger screens.

new-pipeline/style/dashboards.css
```diff
@@ -126,10 +126,15 @@
 .mg-marker-text { font-size: 12px !important; }
 .mg-markers line { stroke-width: 2px !important; }
 
-#dist-caption {
+#dist-caption, #measure-description {
+  margin-left: 2em;
+  white-space: nowrap;
+  width : 55em;
```
georgf: Do we need to limit the width?
It would be great if we could just use the full available page width for the description text.

georgf (commented):
Visually this looks good.
It would be nice though if we can use the full available width for the description text.
Does this work?

new-pipeline/style/dashboards.css
Sylvia23: Okay ! Will do !

new-pipeline/evo.html
Sylvia23: Thanks :D

new-pipeline/src/dashboards.js
Sylvia23: Actually, I did the same initially, but the "More details" option doesn't come without this JavaScript snippet. While,  CSS decreases the length of the long description.

Sylvia23:
@georgf , I have removed the width. Now, we can use the full width.
Can you review the PR.
Thanks!

new-pipeline/src/dashboards.js
Sylvia23: @georgf , Now, I got what you exactly want. I am working on it now. Hopefully, it will be ready in sometime. Actually, It requires very involved code change. I am fixing the bugs that I am getting.
Thanks!

Sylvia23:
@georgf , Can you please review this PR. Thanks!

new-pipeline/src/dist.js
```diff
@@ -886,9 +886,12 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
     var description = histogramsList[0].histograms[0].description
     var metric = histogramsList[0].histograms[0].measure
     var channel = $("#channel-version").val().split("/")[0]
-    var desc = getDescriptionWithLink(metric, channel, description)
+    var desc = getDescription(metric, channel, description)
+    var link = getDescriptionLink(metric, channel, description)
 
     $('#dist-caption').html(desc);
+    $('#caption-dots').html("...");
+    $('#dist-caption-link').html(link);
```
georgf: I think `text-overflow: ellipsis;` should already take care of adding the "...".

new-pipeline/src/dashboards.js
```diff
@@ -1085,6 +1084,16 @@ function getDescriptionWithLink(metric, channel, description) {
   });
   div.append($("<span>").text(descr));
 
+  return div;
+}
+
+function getDescriptionLink(metric, channel, description) {
+  var metricUrl = buildDictionaryURL(metric, channel, description);
+
+  var div = $("<div>", {
```
georgf: Is this `div` element used somewhere?

georgf (commented):
Sorry for missing this before, i didn't see it in my Github notifications.

I think we don't need JavaScript to add the ellipsis, that should be possible with just the `text-overflow` CSS.

Testing this locally, this doesn't cut off to overflow currently (see screenshot).
We should make this first, work out how to make the overflow work from CSS only.
There are some other comments below, but they are not so important right now.

screenshot-2018-3-2 measurement dashboard 1

Sylvia23:
@georgf , I have used CSS for ellipsis. I removed the JS part. This shortening of caption takes place by CSS `text-overflow : ellipsis` only now. JS is used just to dispaly the caption. Now, it will work for screen of any size which was the concern by using JS.
Please have a look!
Thanks!

new-pipeline/src/dist.js
Sylvia23: @georgf , Okay, I will look into it and report ASAP. Thanks!

Sylvia23:
I just saw , the `overflow` is not working propperly. I will fix it now.
Thanks!

new-pipeline/src/dashboards.js
Sylvia23: No, I will remove it. Thanks!

Sylvia23:
@georgf , I have fixed it. Can you please review the PR. Thanks!

new-pipeline/dist.html
```diff
@@ -55,7 +55,11 @@ <h2 style="font-size: 18px; margin-bottom: 0">
                 </h2>
                 <hr></hr>
                 <figure class="col-md-12">
-                  <figcaption id="dist-caption"></figcaption>
+                  <figcaption id="dist-caption" class="overflow ellipsis"></figcaption>
+                  <div class="caption-div">
+                  <figcaption id="caption-dots" >.....</figcaption>
```
georgf: I think this is not working right yet. The "..." should be added automatically through the "text-overflow: ellipsis;" command.
Let's maybe try this isolated without all the existing code around first. I think [this is](https://jsfiddle.net/exzLLez0/11/) roughly the structure that we want here (not pretty, but shows the idea).

Can we apply this idea in this dashboard?

Sylvia23: @georgf ,  I had to add "..." separately because If we put the content(caption) directly into the `HTML` tag `<figcaption></figcaption>`, then `"..."` comes automatically.
But if the caption is rendered using JavaScript, then this `"..."` doesn't come.
Otherwise, the shortening of caption is taking place as per the desired UI.

Please see the screenshot attached.

screenshot from 2018-03-06 02-11-45
screenshot from 2018-03-06 02-11-52

Any advice on what and where to get the changes done is most welcome! 
Thanks!

georgf: I think the ellipsis in the screenshot looks fine. The only issues is that the "more details" button is on a new line.
Am i missing some problem here? :)

Sylvia23: The `ellipsis` looks fine in the screen-shot because I have put an example caption in the `HTML`tag directly.  Please see the code in the screen-shot.

screenshot from 2018-03-06 02-11-45

This code shows up like : 

screenshot from 2018-03-06 02-11-52

The problem comes when the caption is rendered using JavaScript in the runtime from `getDescription` function here [https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/new-pipeline/src/dashboards.js#L1059](url) .
In this case, the caption does shorten up like what a normal `ellipsis` does except `"..."` doesn't come. Hence, I had to add `"..."` separately.

screenshot from 2018-03-06 02-57-23

This shows up like :

screenshot from 2018-03-06 03-00-37

Does this sound clear?

Sylvia23: @georgf , I have found a fix for my problem. I will report back in a while. Thanks!

Sylvia23:
@georgf , Now, the UI looks like : 

For long captions: 
screenshot from 2018-03-06 05-39-20

For short captions:
screenshot from 2018-03-06 05-40-50

One thing is, the caption and "More Details" can't be aligned in one line using `CSS` because `ellipsis` and `inline-block` don't work together as far as I know. One way is to fix the `width` of caption (e.g 80%) but then it will create problem with short captions. 
Any advice with the implementation is most welcomed. Thanks !

georgf:
I did a little research and found that [this jsfiddle](https://jsfiddle.net/y7404myo/3/) seems to do what we want here: with small window widths the text on the left overflows with an ellipsis, but the button on the right stays unaffected.
Can you try that approach?

Sylvia23:
@georgf , Yes, `display : flex` worked for me rather than `display : inline-block`. Thanks for the guidance.
It is good for review now.

georgf:
Great, i will look over it tomorrow!

Can you publish your dashboard through Github pages and comment with the live link here? That makes it much easier to check how the changes look.
You can enable this in your forked repository under "settings", "Github pages".

Sylvia23:
@georgf , I have already done this.
Here is the link : [https://sylvia23.github.io/telemetry-dashboard/new-pipeline/dist.html#!cumulative=0&end_date=2018-03-05&keys=__none__!__none__!__none__&max_channel_version=nightly%252F60&measure=SCALARS_BROWSER.SESSION.RESTORE.BROWSER_STARTUP_PAGE&min_channel_version=nightly%252F57&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=2018-01-22&table=0&trim=1&use_submission_date=0](url)

Sylvia23:
@georgf , I have already done it. Here is the link : [https://sylvia23.github.io/telemetry-dashboard/](url).
Are you able to see this?

georgf:
Yes, i see that, cheers!

One problem i notice on that page:
When i load it it looks fine. But when i reduce the window width, suddenly the description is shown multiple times:
screen shot 2018-03-07 at 14 20 21

Checking this in the inspector, it looks like the text is actually added to the page multiple times:
screen shot 2018-03-07 at 14 22 36

Sylvia23:
@georgf , Should we do like , we reload the page whenever the user resize the window to fix this issue!?
Please suggest. Thanks!

georgf:
@Sylvia23 I think the problem is in the JavaScript changes here. It looks like we are appending the text repeatedly instead of replacing it.
This is a new problem, so it is probably a problem in the JavaScript that you changed.
Does that help tracking the problem down?

Sylvia23:
@georgf , I have fixed it. Actually I changed the `$('dist-caption').html(desc)` to `$('dist-caption').innerHTML += desc` due to which it was appending the caption everytime the function gets called. 
Now, it's good to go. Please review. Thanks!

new-pipeline/dist.html
```diff
@@ -55,7 +55,10 @@ <h2 style="font-size: 18px; margin-bottom: 0">
                 </h2>
                 <hr></hr>
                 <figure class="col-md-12">
-                  <figcaption id="dist-caption"></figcaption>
+                <div class="caption-div">
+                  <figcaption id="dist-caption" class="overflow ellipsis"></figcaption>
```
georgf: A small request: let's call this `dist-caption-text`, to better tell the two elements apart.

new-pipeline/evo.html
```diff
@@ -54,6 +54,13 @@ <h2 style="font-size: 18px; margin-bottom: 0">
                     as
                     <select id="filter-process-type" class="multiselect" title="Process type" data-all-selected="any process type"></select>
                 </h2>
+                <hr></hr>
+                <div class="caption-div">
+                <figcaption id="evo-caption" class="overflow ellipsis"></figcaption>
```
georgf: Likewise, lets use `evo-caption`. This and the following element are missing indentation (should be 2 spaces per level).

new-pipeline/evo.html
```diff
@@ -54,6 +54,13 @@ <h2 style="font-size: 18px; margin-bottom: 0">
...
+                <figcaption id="evo-caption-link"></figcaption>
+                </div> 
+                <h2 style="font-size: 18px; margin-bottom: 0">Showing graphs for <select id="selected-key" class="multiselect" title="Selected key"></select></h2>
```
georgf: This is duplicated, you can see [here](https://bit.ly/2FBu2ac) how it shows up two times.

new-pipeline/src/dist.js
```diff
@@ -874,9 +874,11 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
     var description = histogramsList[0].histograms[0].description
     var metric = histogramsList[0].histograms[0].measure
     var channel = $("#channel-version").val().split("/")[0]
-    var desc = getDescriptionWithLink(metric, channel, description)
-
+    var desc = getDescription(metric, channel, description)
+    var link = getDescriptionLink(metric, channel, description)
+    var div = document.getElementById('dist-caption');
```
georgf: This is more clearly structured now, nice!

georgf (commented):
This looks good now, cheers!
I just have some comments below on smaller cleanup, then this should be done.

We should improve the text styling later, but that is not important here. #406 will address that part.

new-pipeline/dist.html
Sylvia23: Sure!

new-pipeline/src/dist.js
Sylvia23: Yeah! 😀

new-pipeline/evo.html
Sylvia23: Here, you mean `evo-caption-text` Right?

new-pipeline/evo.html
Sylvia23: Yeah, I see.

Sylvia23:
@georgf commited the changes requested. Ready for final review. Thanks!

new-pipeline/dist.html
```diff
@@ -55,7 +55,10 @@ <h2 style="font-size: 18px; margin-bottom: 0">
                 </h2>
                 <hr></hr>
                 <figure class="col-md-12">
-                  <figcaption id="dist-caption"></figcaption>
+                <div class="caption-div">
```
georgf: This `div` is missing one indentation level. Then the `figcaption` need to move along accordingly.

new-pipeline/dist.html
```diff
@@ -55,7 +55,10 @@ <h2 style="font-size: 18px; margin-bottom: 0">
...
+                  <figcaption id="dist-caption-text" class="overflow ellipsis"></figcaption>
+                  <figcaption id="dist-caption-link"></figcaption>
+                </div></br></br>
```
georgf: I think we should not use `br` here. Instead lets use `margin-bottom` or similar.

new-pipeline/src/dist.js
```diff
@@ -874,11 +874,13 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
     var description = histogramsList[0].histograms[0].description
     var metric = histogramsList[0].histograms[0].measure
     var channel = $("#channel-version").val().split("/")[0]
-    var desc = getDescriptionWithLink(metric, channel, description)
-
-    $('#dist-caption').html(desc);
+    var desc = getDescription(metric, channel, description)
+    var link = getDescriptionLink(metric, channel, description)
```
georgf: These lines are missing a semicolon at the end.

new-pipeline/src/evo.js
```diff
@@ -324,11 +324,12 @@ $(function () {
 
               var metric = $("#measure").val()
               var channel = $("#max-channel-version").val().split("/")[0]
-              var description = getDescriptionWithLink(metric, channel, evolutionDescription)
-
+              var description = getDescription(metric, channel, evolutionDescription)
+              var link = getDescriptionLink(metric, channel, description)
```
georgf: These lines are also missing semicolons.

georgf (commented):
Thanks! There are still a few small things i noticed, but visually this is done :)

Sylvia23:
@georgf pushed the changes requested. Thanks!

new-pipeline/dist.html
```diff
@@ -55,7 +55,10 @@ <h2 style="font-size: 18px; margin-bottom: 0">
                 </h2>
                 <hr></hr>
                 <figure class="col-md-12">
-                  <figcaption id="dist-caption"></figcaption>
+	                <div class="caption-div">
+	                  <figcaption id="dist-caption-text" class="overflow ellipsis"></figcaption>
+	                  <figcaption id="dist-caption-link"></figcaption>
+	                </div>
                     <div class="col-md-9" id="plots">
```
georgf: The indentation is still a bit off here.
The goal would be to have `dist-caption` on the same indentation level as `plots` (as they are at the same level in the hierarchy here).

georgf (commented):
Thanks - one last small change needed, then we're good!

new-pipeline/dist.html
Sylvia23: @georgf , I put `dist-caption` on the same level as that of `plots` and again pushed. But it still shows it like this, they are on the same level in my code on my local machine. Can you please review and tell the probable reasons. Thanks!

georgf: Are you mixing tabs and spaces here?
Most code editors have an option to show these with extra characters.
In the code in this repository, we only use spaces for indentation, not tabs.

Sylvia23: @georgf , This is how the code looks like in my editor. 

screenshot from 2018-03-10 13-17-39

Please have a look. Thanks!

georgf: If you look in the ["files changed" view](https://github.com/mozilla/telemetry-dashboard/pull/418/files) you can see that it looks different.

To fix this, you can check for an option to show tabs & spaces in your editor, so you can see this.
More generally you can check for an option in your editor that let's you choose to use spaces instead of tabs for indentation.

Sylvia23: @georgf , Yeah I can see the difference when I push the branch on GitHub. But it is not visible in my editor. I already have `Indent using spaces` setting in my editor. I will try again and hope to fix it. Thanks!

Sylvia23:
@georgf , I think it's fixed now. We are now good to go. Thank you so much!

Sylvia23:
@georgf , Can you please review this whenever you are free. Thanks!

new-pipeline/dist.html
```diff
@@ -55,8 +55,11 @@ <h2 style="font-size: 18px; margin-bottom: 0">
                 </h2>
                 <hr></hr>
                 <figure class="col-md-12">
-                  <figcaption id="dist-caption"></figcaption>
-                    <div class="col-md-9" id="plots">
+                        <div class="caption-div">
+                            <figcaption id="dist-caption-text" class="overflow ellipsis"></figcaption>
```
chutten: It doesn't matter to browsers, but it might interest you to know that the spec expects `figcaption` elements to be either the first or last child (or one of each) of a `figure` element. [ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure)

new-pipeline/src/dist.js
```diff
@@ -874,11 +874,13 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
...
-
-    $('#dist-caption').html(desc);
+    var desc = getDescription(metric, channel, description);
+    var link = getDescriptionLink(metric, channel, description);
+    var div = document.getElementById('dist-caption-text');
+    $('#dist-caption-text').html(desc);
+    $('#dist-caption-link').html(link);
   } else {
-    $('#dist-caption').text(""); // Clear the histogram caption
+    $('#dist-caption-text').text(""); // Clear the histogram caption
```
chutten, line 882: Should we also clear the link? (it is possible this case is never reached, so it may not matter except for symmetry)

new-pipeline/style/dashboards.css
```diff
@@ -126,10 +126,27 @@
...
 
-#dist-caption {
-  font-size: 1.5em;
+.caption-div {
+  display: flex;
+  justify-content: center;
+  align-items: center;
   text-align: center;
-  margin: 0.1em;
+  margin: 0 auto;
```
chutten: `.caption-div` has full width, so automatic horizontal margins like these will always calculate to 0. I think we can omit this.

new-pipeline/evo.html
```diff
@@ -54,7 +54,11 @@ <h2 style="font-size: 18px; margin-bottom: 0">
                     as
                     <select id="filter-process-type" class="multiselect" title="Process type" data-all-selected="any process type"></select>
                 </h2>
-                <h3 id="measure-description" style="font-size: 14px; margin: 0 3px"></h3>
+                <hr></hr>
+                <div class="caption-div">
+                    <figcaption id="evo-caption-text" class="overflow ellipsis"></figcaption>
```
chutten: I think the `.ellipsis` class may be holding on from an earlier version of this patch. It looks like `.overflow` takes care of it now, so we can remove it here and in `dist.html`.

new-pipeline/style/dashboards.css
```diff
@@ -126,10 +126,27 @@
...
+  margin-bottom: 3em;
+}
+
+.overflow {
+    white-space: nowrap !important;
+    overflow: hidden !important;
+    text-overflow: ellipsis !important;
```
chutten: `!important` used to be (may still be? It's been a few years since I've asked) only used if absolutely necessary. I think we can omit them here, as no other classes are trying to set `overflow`, `text-overflow`, or `white-space` on elements with `.overflow`

new-pipeline/dist.html
```diff
@@ -55,8 +55,11 @@ <h2 style="font-size: 18px; margin-bottom: 0">
...
+                            <figcaption id="dist-caption-link"></figcaption>
+                        </div>
+                        <div class="col-md-9" id="plots">
```
georgf: This looks like the spaces are sorted now!
Now we just need to indent `dist-caption` and `col-md-9` one level less, as they're not on the same level as the following `<div class="row">`.

chutten (commented):
I'm not @georgf but I'm helping out (he's going to be on vacation soon, so I'm picking things up where I can) : )

I have a couple of questions, a couple of comments, and some CSS opinions. At this point I've tested the changes and the behave perfectly well, so these are minor code cleanups. If you feel as

new-pipeline/evo.html
Sylvia23: I removed it from `dist.html` earlier. I could not see it there now. Please correct me if I am missing something.

Sylvia23: Ohh I got it I guess. You are saying that I should keep the class name as `overflow` rather than `overflow ellipsis`. Right?

chutten: Correct!

Sylvia23:
@chutten @georgf , I have made the suggested changes. Please have a look. Thanks!

new-pipeline/style/dashboards.css
```diff
@@ -126,10 +126,26 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
   text-align: center;
-  margin: 0.1em;
+  margin-bottom: 3em;
+}
+
+.overflow {
+    white-space: nowrap;
```
chutten: This file uses two-space indents. This selector's block is at four

chutten (commented):
Argh, I just noticed the CSS indentation. I'm sorry I didn't notice it last time.

chutten:
We _really_ need a linter to catch these things on developer machines. The prototype will definitely have one.

Sylvia23:
@chutten , Yeah, linter automatically analyses the code for these errors. It's a kind of must for this.

I have corrected the indentation. Please see. Thanks!

new-pipeline/src/dist.js
```diff
@@ -874,11 +874,14 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
     var description = histogramsList[0].histograms[0].description
     var metric = histogramsList[0].histograms[0].measure
     var channel = $("#channel-version").val().split("/")[0]
-    var desc = getDescriptionWithLink(metric, channel, description)
-
-    $('#dist-caption').html(desc);
+    var desc = getDescription(metric, channel, description);
+    var link = getDescriptionLink(metric, channel, description);
+    var div = document.getElementById('dist-caption-text');
```
chutten: What is div used for?

new-pipeline/style/dashboards.css
```diff
@@ -126,10 +126,26 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
   text-align: center;
-  margin: 0.1em;
+  margin-bottom: 3em;
+}
+
+.overflow {
+	white-space: nowrap;
```
chutten: The whitespace appears to still be off. It looks like it's a tab character instead of spaces?

new-pipeline/src/dist.js
Sylvia23: We can drop it. I used it initially for something. But now its not needed. I have removed it.

Sylvia23:
@chutten , I have removed the white space. Does they still appear? Please see. Thanks!

chutten (approved):
Looks good now!

PR #424, opened 2018-03-01
Fix #423: Move permalink generator to front of button row

### LINKED ISSUE #423: Move short link button to a more discoverable location
The short link button in the top right corner is hard to discover for people.
This probably affects `dist.html` and `evo.html` in `/new-pipeline`.

We would like to try to move it elsewhere and see if that improves discoverability.

screen shot 2018-02-28 at 14 38 41

ecomerford:
@georgf , I'm working on a mock-up of more changes, but here is a quick update that moves the permalink generator to the front of the button row. Please review! Thanks :)

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -14,7 +14,6 @@
     <link rel="stylesheet" type="text/css" href="lib/daterangepicker-bs3.css" />
     <link rel="stylesheet" type="text/css" href="style/elessar.css" />
     <link rel="stylesheet" type="text/css" href="style/metricsgraphics.css" />
-
     <link rel="stylesheet" type="text/css" href="style/dashboards.css" />
     <script async src="https://www.googletagmanager.com/gtag/js?id=UA-49796218-2"></script>
 </head>
@@ -22,11 +21,11 @@
     <div class="container-fluid">
         <header>
             <form class="navbar-form inline pull-right">
+                <span class="permalink-control"></span>
                 <a href="https://github.com/mozilla/telemetry-dashboard/issues" class="btn btn-success" title="Report a bug in Telemetry dashboards" target="_blank">Report bug</a>
                 <a href="./evo.html" class="btn btn-default" id="switch-views" title="Switch to the evolution view">Evolution View</a>
                 <a href="/" class="btn btn-default btn-primary" title="Landing page for Telemetry dashboards">Dashboards Home</a>
                 <a href="./tutorial.html#HistogramDashboard" class="btn btn-default" id="tutorial" title="Usage tutorial for Telemetry dashboards">Usage Tutorial</a>
-                <span class="permalink-control"></span>
             </form>
             <h1>Measurement Dashboard</h1>
             <div class="error-msg">
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -4,9 +4,7 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-
     <title>Measurement Dashboard</title>
-
     <!-- Bootstrap and supporting libraries -->
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css">
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -20,8 +20,8 @@
 <body>
     <div class="container-fluid">
         <header>
-            <form class="navbar-form inline pull-right">
-                <span class="permalink-control"></span>
+            <form class="navbar-form inline pull-right btn-group">
+                <span class="permalink-control btn-default">Get Permalink</span>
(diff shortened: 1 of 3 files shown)


georgf:
This addresses #423.

new-pipeline/evo.html
```diff
@@ -4,27 +4,27 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    
+
```
georgf: Small thing: There are some unrelated whitespace changes in this change, possibly from auto-formatting from your editor?

new-pipeline/dist.html
```diff
@@ -14,19 +14,18 @@
...
     <link rel="stylesheet" type="text/css" href="style/metricsgraphics.css" />
-
     <link rel="stylesheet" type="text/css" href="style/dashboards.css" />
     <script async src="https://www.googletagmanager.com/gtag/js?id=UA-49796218-2"></script>
 </head>
 <body>
     <div class="container-fluid">
         <header>
             <form class="navbar-form inline pull-right">
+                <span class="permalink-control"></span>
```
georgf: Great, that already makes it better!
Is it easy to add a title to it too in this PR?
It looks like this will require diving into JS in dashboards.js.

georgf (commented):
Great start!
I'd love to see a text on the button too if it's not hard to add.

georgf:
This is how it looks for me with the current change:

screen shot 2018-03-01 at 12 48 02

georgf:
Following up to a comment from #423:
> I wanted to give it a name to draw more attention to it, but I can't figure out a good way to do so. I tried turning it into a button but that became an awkward button inside of a button. Plus, the generator made it larger than the other buttons in its row. I can't seem to find any other Bootstrap class types that would do the trick.
> 
> I suppose I could make it into a button, then disable it and try to adjust the CSS to make it the same size as the others, but that seems a bit strange.

I can't currently visualize the problem, could you share that?
I'm not sure if it fixes your problem, but i wonder if we could give the button a title, then hide the title when we show a link?

ecomerford:
@georgf Gotcha, I'll try looking into the JS to add a title tonight. And remove the white space in my code! If I'm still having trouble, I'll go ahead and share a screen shot to show you what I meant re: making the generator into a button.

Thank you for the feedback! More coming soon!

ecomerford:
@georgf @chutten here's what I meant by "button within a button."
screenshot - 03012018 - 06 36 30 pm

ecomerford:
In this screenshot, though, I've taken out the `append` function within the permalink-control JS and replaced it with the class of button, name "Get Permalink" within the HTML. I don't think it removed any of the functionality as the button JS is all still there. Let me know what you think!

P.S. I also made them a button group, as you can see... I thought it looked a little nicer but happy to change it back if you don't like it!
screenshot - 03012018 - 06 38 12 pm

georgf:
Visually that button group makes sense to me and seems like a good improvement!
@chutten, what do you think?

georgf (commented):
This looks like a good approach to me, it actually makes things cleaner and the UI more clear!

I took it for a test drive [over here](https://georgf.github.io/telemetry-dashboard/new-pipeline/dist.html) and it seems the permalink button does not work currently?

ecomerford:
@georgf I noticed that this morning, as well. I'm going to have to keep playing around with it! I think I removed too much on the JS side...

chutten:
The click listener is added in dashbaords.js around line 138 (  ` $(".permalink-control button").click` ), in case that helps.

I'm not as wild about the button group. The actions the buttons take are very different, so bringing them "closer" in the design feels odd to me. But that's just me : )

I like the text "Get Permalink" that's very easy to spot, and very clear about what it does. It also echoes how others of mozilla's tools (like searchfox)  talk about this sort of thing, which is beneficial

ecomerford:
@chutten thanks, Chris! I'll take a look at the click listener. 

I see what you mean about the buttons being really different. I was thinking that the grouping just draws attention to all of the  buttons, which seems beneficial. Plus I just like the way it looks :p but I can definitely change it back!

On a similar note, would the "Report A Bug" button make more sense as a `btn-danger`, or is it green to encourage people to report issues?

georgf:
> I see what you mean about the buttons being really different. I was thinking that the grouping just draws attention to all of the buttons, which seems beneficial.

I agree with this, this seems to help a lot with scanning the page.
Checking back with Chris quickly, let's move ahead with your proposal Erin! :)

So the main blocker here now is to make the button work.

chutten:
I'm not sure what the design intent was behind the choice of green for that button. It's been that way ever since I started working on this project. Let's leave it that way for now.

ecomerford:
I think this has fixed it. It's hard to tell whether the button is working because I've been using a local server, and I don't think the bitly URL works on that. But I'm not 100% sure. At this point, I haven't removed anything from the JS, only added the "Get Permalink" name.

georgf (commented):
I tried to apply the changes locally to test them, but GitHub fails to apply them. At the top of this PR it says that the commits come from an "unknown repository".

Did you happen to delete your forked repository? Or make it private?
Feel free to just submit a new PR if that is easier.

georgf:
One request: When you do the PR from your own fork, can you publish a live view of the web page?
You can just go to your repository, to "settings", then "GitHub Pages" to publish it, then let us know the link.
That makes it easy for us to test & review the changes.

PR #425, opened 2018-03-01
Fix #400 : Rename Aurora channel to DevEdition in the front-end

### LINKED ISSUE #400: Rename Aurora channel to DevEdition in the front-end
We still show aurora in multiple places.
My understanding is that this is the DevEdition data. Is that correct?

If yes, can we rename this in the UI to be "DevEdition" or similar, to make it clear to users?

Sylvia23:
"aurora" name replaced with "dev edition" in UI according to the data.
The presence of the name in the repo can be easily found with the git command `git grep "aurora"`.
Hence, knowledge of git can easily solve the issue. Two files 

> evo.js

> dashboards.js

 have been affected.

### DIFF
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -714,12 +714,14 @@ function getHumanReadableOptions(filterName, options) {
       options = options.concat([null])
         .concat(badOptions);
     }
+
     return options.map(function (option) {
-      return option !== null ? [option, option.replace("/", " ")] : null;
+      return option !== null ? [option, option.replace("aurora", "dev edition").replace("/", " ")] : null;
     });
+		   
   }
   return options.map(function (option) {
-    return [option, option]
+    return [option, option];
   });
 }
 


new-pipeline/src/dashboards.js
```diff
@@ -433,7 +433,7 @@ function getFilterSetsMapping(filters, comparisonName) {
 function getHumanReadableOptions(filterName, options) {
   var channelVersionOrder = {
     "nightly": 0,
-    "aurora": 1,
+    "dev edition": 1,
```
georgf: I see, now i understand the problem. This doesn't actually change the way this is displayed, we don't need to change this i think.

The interesting part is near the end of this function, [around here](https://github.com/mozilla/telemetry-dashboard/blob/ead6c6705de571c45fb3be27f16f0a2b29d8f164/new-pipeline/src/dashboards.js#L718).
You could change this to `console.log()` the value of what's returned (or inspect it in the debugger?), then you'll see how this looks and can change it's values.

Sylvia23: @georgf , I have looked into the code, I have figured out from where the names can be replaced. 
But I am a little confused with the implementation.

The `map` function returns "arrays of length 2 or boolean" in the one-liner function on line 

[https://github.com/mozilla/telemetry-dashboard/blob/ead6c6705de571c45fb3be27f16f0a2b29d8f164/new-pipeline/src/dashboards.js#L718](url)

I was trying to expand the one-liner syntax as : 

`return options.map(function (option) {
        if (option) {
            option == option.replace("/", " ");
            return option;
        }
        else {
            return option == null;
        }
    });`

But it is throwing an error in the end. The error says : 

`uncaught exception: Bad options value: must be an array of arrays, either each with two strings or each with three strings`

screenshot from 2018-03-05 04-41-49

Hence, I am wondering if both the functions (one-liner and the expanded) are the same.

Can you please help me with this ? Thanks!

chutten: I'm not @georgf, but I feel like I might be able to help anyway : )

Your expanded syntax (with some whitespace for clarity):
```js
return options.map(function (option) {
  if (option) {
    option == option.replace("/", " ");
    return option;
  } else {
    return option == null;
  } });
```

That first line within the `if (option)` block has two equal signs. It will evaluate to a boolean, and then be discarded. I think you only want a single equal sign.

You could leave this line alone and take the functional programming route and perform a map before this one. Something like

```js
options = options.map(option => /* something about replacing 'aurora' with 'dev edition' */)
return options.map(...this code that already exists);
```

Is this helpful?

Sylvia23: @chutten , Thank you so much for the help. Actually, first of all I was doing the same thing but caught up in some error. Now, I have fixed it.

Sylvia23:
@chutten @georgf , Can you please review the PR. I have changed the channel name "Aurora" to "dev edition" in the UI.
Thanks!

new-pipeline/src/dashboards.js
```diff
@@ -714,12 +714,26 @@ function getHumanReadableOptions(filterName, options) {
...
+	      newArray.push(actual[i]);
+	    }
+	  }
+	  return newArray;
+	}
+	options = cleanArray(options);
+
+    options = options.map(option => option.replace("aurora", "dev edition"));
+	return options.map(function (option) {
       return option !== null ? [option, option.replace("/", " ")] : null;
```
georgf: This is addressing things at the right spot, but it's not quite functional yet.
Testing this locally i see that the line separators are missing and clicking the entries in the list doesn't show any data.
I assume this is because now we are returning a list with entries like `["dev edition", "dev edition"]` instead of the expected `["aurora/58", "dev edition"]` or so.

I think we need to leave out the first `...map()`.
Then the second `map()` should return something like `... ? [option, niceName] ...` (where `niceName` is the one that would have "dev edition".
Then take it for a local test and see if it works.

Sylvia23:
@chutten @georgf , I have taken the local test and it works properly. Can you please review this. 
Here is the live link : [sylvia23.github.io/telemetry-dashboard/new-pipeline/dist.html](url)
Thanks!

chutten (approved):
It behaves well, the code is clear. Let's get this live.

chutten:
Thank you for your contribution!

Issue #426, opened 2018-03-01
Add icons for each section on landing page


georgf:
To aid visually guided people, we should add an icon to each main section on [the landing page](https://mozilla.github.io/telemetry-dashboard/):

<img width="400" alt="screen shot 2018-03-01 at 18 14 53" src="https://user-images.githubusercontent.com/407696/36858895-0d6cecd8-1d7d-11e8-9a91-b7c07e7ed2e1.png">

georgf:
Shreya will probably work on this.

georgf:
This will involve changing `/index.html`.
I propose we use font-awesome icons for this:
- ["chart-bar"](https://fontawesome.com/icons/chart-bar?style=solid) for main tools
- ["book"](https://fontawesome.com/icons/book?style=solid) for documentation
- ["question-circle"](https://fontawesome.com/icons/question-circle?style=solid) for "getting help"
- ["wrench"](https://fontawesome.com/icons/wrench?style=solid) for "other tools"

font-awesome icons are added by adding a `<i class="$iconName"></i>`, e.g. `<i class="fas fa-wrench"></i>`.

shreya99oak:
Yes, sure!

manishakeim:
Is this issue assigned to anyone ? or Can I claim it ?

georgf:
This was already fixed in #433, we just forgot to close this issue.
If you can't find any other open issues though, let us know, we'll find something!

manishakeim:
@georgf Can you please help me find one, as I am a beginner I was finding more of 'good first issue' and there are no such issues with these labels left how should I proceed with contributing to this project for Outreachy?

georgf:
Hi @manishakeim, sure! How about #469?
I should call out that there are active applications already on this Outreachy project, while others are still looking for applications. Maybe take a look on other projects as well :)

PR #433, opened 2018-03-03
Add icons for each section on landing page #426

shreya99oak:
I added the font-awesome icons to the headings, pls review it!

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,8 +4,6 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <script defer src="https://use.fontawesome.com/releases/v5.0.8/js/all.js"></script>
-
     <title>Telemetry portal</title>
 
     <!-- Bootstrap and supporting libraries -->
@@ -29,7 +27,7 @@
         </header>
         <div class="row">
             <div class="col-md-4">
-                 <h3><i class="fas fa-chart-bar"></i> Main tools</h3>
+                <h3><i class="fa fa-bar-chart" aria-hidden="true"></i> Main tools</h3>
                 <div class="dashboard-list list-group" style="margin: 0">
                     <a class="list-group-item" href="new-pipeline/dist.html">
                         Measurement Dashboard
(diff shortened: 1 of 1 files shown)


chutten:
Do you happen to have a live page with this patch? I notice you have this pushed to the gh-pages branch of your fork, is that public? Or can you provide a screenshot, so I can take a look?

shreya99oak:
icons

index.html
```diff
@@ -4,6 +4,7 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <script defer src="https://use.fontawesome.com/releases/v5.0.8/js/all.js"></script>
```
chutten: We already include the FA CSS in the page. Do we actually need the JS?

chutten (changes requested):
The screenshot looks good, thanks!

index.html
shreya99oak: Hi, I tried it without the JS, but it displays empty boxes when you open it on the browser. I'll attach a screenshot..

shreya99oak: without_js

georgf: Going from [the documentation here](https://fontawesome.com/how-to-use/web-fonts-with-css), it should be possible to use font awesome with CSS only.
I might have given the wrong names in #426. We currently use font-awesome 4 on the landing page, but the links are for the font-awesome 5 names. You can try out [the names for fa 5](https://fontawesome.com/v4.7.0/icon/bar-chart/) instead.

shreya99oak: Oh, okay

index.html
```diff
@@ -171,4 +171,4 @@ <h3>What's new</h3>
     <script src="analytics/analytics.js"></script>
     <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>
 </body>
-</html>
+</html
```
georgf: This change looks unintentional? Lets remove it.

georgf (commented):
This looks good, thanks!
There is one small change needed, then this is good to go.

index.html
shreya99oak: oh sorry,I'll change that

georgf (approved):
Thanks, this is a nice improvement!

shreya99oak:
Thanks for the help:) 
Could u pls suggest some bugs to work on ?

PR #435, opened 2018-03-05
Add icons for each section on landing page - Issue #426 

paucaro:
I added the font-awesome icons to the headings sections.
- The project use fontawesome of bootstrap in a version 4.3.0 and the link of the task proposes the icons to use are fontawesome version 5, so I am using the equivalents in the version that is being used.
- I proposed the newspaper icon 'fa fa-newspaper-o' for the What's new section, because I think it represents the news of a page
- I thought that to give Mozilla more visibility, it could use the colors of this one; so change the color of the icon (as a proposal) to a gradient with the colors that I found on this page https://www.mozilla.org/en-US/styleguide/identity/marketplace/color/

I attached an image with the modifications made. Please review it :)
image

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
+
 <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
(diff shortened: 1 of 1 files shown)


georgf:
Sorry, but icons were already added in #433 :)
Per email i suggested you another possible improvement to work on, looking forward to see that!

PR #436, opened 2018-03-06
Fix Issue #423: Add name to permalink generator, move to front of button row.

ecomerford:
Hi @georgf @chutten, here I've moved the permalink generator to the front of the button row, and added a name to it. Plus, the short-link generator still works!

Let me know what you think. Thank you!

You can view it live here: https://ecomerford.github.io/telemetry-dashboard/

### DIFF
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -130,7 +130,7 @@ $(document)
     $(".permalink-control")
       .append(
         '<div class="input-group">' +
-        '    <span class="input-group-btn"><button type="button" class="btn btn-default" title="Get Permalink"><span class="glyphicon glyphicon-link"></span></button></span>' +
+        '    <span class="input-group-btn"><button type="button" class="btn btn-default" title="Get Permalink">Get Permalink<span class="glyphicon glyphicon-link"></span></button></span>' +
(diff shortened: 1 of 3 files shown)


new-pipeline/src/dashboards.js
```diff
@@ -130,7 +130,7 @@ $(document)
     $(".permalink-control")
       .append(
         '<div class="input-group">' +
-        '    <span class="input-group-btn"><button type="button" class="btn btn-default" title="Get Permalink"><span class="glyphicon glyphicon-link"></span></button></span>' +
+        '    <span class="input-group-btn"><button type="button" class="btn btn-default" title="Get Permalink ">Get Permalink</button></span>' +
```
georgf: A small thing here, do we need the trailing space in `title="Get Permalink "` ?

georgf (commented):
This looks good to me, thanks!
I only have a small thing below before we merge this.

The change to a button group you had before was a good improvement i think. Would you be interested to do this in a follow-up issue?

ecomerford:
Sorry about that, it's fixed now! I'd be happy to work on the button group again in another issue.

georgf (approved):
Thanks!

Issue #251, opened 2016-10-06
Some API parameters are not defined correctly


longaraymatheus:
Looking at the [documentation](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/wrapper/README.md) (https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/wrapper/README.md), _useSubmissionDate_, _sanitize_, _trim_ and _sensibleCompare_ are defined as booleans parameters.

However, when I called _window.TelemetryWrapper.go(params, parentEl)_, these parameters did not change the output chart at all. More precisely, _sanitize_, _trim_ and _sensibleCompare_ are the ones I had trouble with.

Going through the code I noticed that some of them are treated as string, not boolean. You can see [here](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/wrapper/telemetry-wrapper.js#L789) (https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/wrapper/telemetry-wrapper.js#L789) that _sanitize_ value is being compared to _string 'false'_, not to _boolean false_ as expected. The same occurs to _trim_ and _sensibleCompare_. _useSubmissionDate_ works as expected.

Please let me know if there is anything I can help with.

chutten:
L789 is a type-coercive comparison, so fret not about types.

Can you provide the params you were using, and a description of what you expected them to produce?

longaraymatheus:
The parameters were the following:

_var plots = [
  {
    "channel": "release",
    "version": "49",
    "metric": "PRINT_PREVIEW_OPENED_COUNT",
    "useSubmissionDate": false,
    "sanitize": false,
    "trim": false,
    "compare": "",
    "sensibleCompare": false,
    "evoVersions": 0,
    "filters": {}
  }
];_

What I expected them to produce was a not-sanitized and not-trimmed  chart. However, a sanitized and trimmed chart was being returned.

Then, I checked https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/wrapper/telemetry-wrapper.js#L760 and changed the parameters to the following. Note that _useSubmissionDate_ works as expected passing it as boolean.

_var plots = [{
  "channel": "release",
  "version": "49",
  "metric": "PRINT_PREVIEW_OPENED_COUNT",
  "useSubmissionDate": false,
  "sanitize": "false",
  "trim": "false",
  "compare": "",
  "sensibleCompare": "false",
  "evoVersions": 0,
  "filters": {}
[...]

PR #439, opened 2018-03-06
Fixed-issue-251-Some API parameters are not defined correctly

Shruti9520:
Some API parameters are not defined correctly.
Fixes- [#251](https://github.com/mozilla/telemetry-dashboard/issues/251)

Files changed-
1. `telemetry-wrapper.js`

### DIFF
--- a/wrapper/telemetry-wrapper.js
+++ b/wrapper/telemetry-wrapper.js
@@ -818,14 +818,14 @@ window.TelemetryWrapper.go = function (params, element) {
       params.filters = {};
     }
     params.useSubmissionDate = params.useSubmissionDate || false;
-    params.sanitize = params.sanitize != 'false';
-    params.trim = params.trim != 'false';
+    params.sanitize = params.sanitize != false;
+    params.trim = params.trim != false;
     params.compare = params.compare; // default undefined
     if (params.compare && params.filters[params.compare]) {
       // If we're filtering to a particular value, we can't then compare by it.
       delete params.compare;
     }
-    params.sensibleCompare = params.sensibleCompare != 'false';
+    params.sensibleCompare = params.sensibleCompare != false;
     params.keyLimit = window.parseInt(params.keyLimit) || 4;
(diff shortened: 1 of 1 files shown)


chutten (commented):
Does this change still retain support for existing string parameters? For instance, if `sanitize = 'false'`, then in the previous version of the code, `params.sanitize` would end up as `false`.

Does that still happen?

Shruti9520:
@chutten Hi

`params.sanitize` is used in places like  [https://github.com/Shruti9520/telemetry-dashboard/blob/412d75899184643e75f4bbd3c794bdedd3b73d88/wrapper/telemetry-wrapper.js#L213](url). If we use `sanitize` as `boolean`, it would return `0` or `1` which is completely fine.

Is it what you are asking? 

Thanks! :)

chutten:
I mean when someone uses telemetry wrapper (see examples in the documentation: https://github.com/mozilla/telemetry-dashboard/tree/gh-pages/wrapper) and passes in `sanitize: "false"` as part of their params, it will be coerced to `true` in your code, won't it?

chutten:
Hrm, actually the documentation _doesn't_ have an example. Here's an example:

```js
TelemetryWrapper.go({
  sanitize: "false",
}, element);
```

Shruti9520:
@chutten Yes, you are right. My suggested approach doesn't seem good. :(

Anyways, I will try some other approach. :)

Thanks!

chutten:
@Shruti9520 How are things on this PR? Anything I can help with?

Shruti9520:
Hi @chutten 
I am sorry for being inactive these days. Will you mind providing me some pointers regarding this?

chutten:
For testing this change, I recommend writing a web page that uses TelemetryWrapper (see its documentation for an example) and then using TelemetryWrapper.go and a debugger to test the various values of true/"true"  and false/"false" to make sure they coerce to the correct values when passed deeper in.

chutten:
Closing due to inactivity. We can reopen this if it becomes active again in future.

PR #453, opened 2018-03-08
Fixes #446: Make probe dictionary header links more discoverable

### LINKED ISSUE #446: Make probe dictionary header links more discoverable
> It's hard to find the "Find probes" and "Stats" links at the top of the page. We should make them more discoverable (e.g. underline them).

(this comes from mozilla/probe-dictionary#15)

JQuinnie:
Fixes #446  

From recently added 'Probe Dictionary' page...
Header links were hard to find on the main index page, CSS styling updates to make the font size larger and add a box overlay to each link, see _Before_ and _After_ below;

**_Before_**
telemetry1before

**_After_**
telemetry1after

Let me know if you would like any additional adders.
[Live site for reference](https://jquinnie.github.io/telemetry-dashboard/probe-dictionary)

### DIFF
--- a/probe-dictionary/explorer.css
+++ b/probe-dictionary/explorer.css
@@ -127,8 +127,3 @@ body {
     font-size: larger;
     padding-right: 8px;
 }
-
-.nav-item > a {
-    background-color: #5496cf;
-    border-radius: 4px;
-}
--- a/probe-dictionary/explorer.css
+++ b/probe-dictionary/explorer.css
@@ -124,6 +124,5 @@ body {
 }
 
 .nav-item {
-    font-size: larger;
-    padding-right: 8px;
+    font-weight: bold;
 }
--- a/probe-dictionary/explorer.css
+++ b/probe-dictionary/explorer.css
@@ -122,3 +122,13 @@ body {
 .hidden {
     display: none;
 }
+
+.nav-item {
+    font-size: larger;
+    padding-right: 8px;
+}
+
+.nav-item > a {
+    background-color: #5496cf;
+    border-radius: 4px;
+}
--- a/probe-dictionary/index.html
+++ b/probe-dictionary/index.html
@@ -31,11 +31,11 @@
       <button class="navbar-toggler navbar-toggler-right" type="button" data-toggle="collapse" data-target="#navbarCollapse" aria-controls="navbarCollapse" aria-expanded="false" aria-label="Toggle navigation">
         <span class="navbar-toggler-icon"></span>
       </button>
-      <a class="navbar-brand" href="#">Probe dictionary</a>
+      <a class="navbar-brand" href="#">Probe Dictionary</a>
 
       <div class="collapse navbar-collapse" id="navbarCollapse">
         <ul class="navbar-nav mr-auto">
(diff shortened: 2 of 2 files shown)


georgf (commented):
Thanks!
For me this looks a bit too large and the different background color distracts me from the main content.

Balancing this is a little harder :) Could we try in comparison with the normal font-size and a bold style?

JQuinnie:
@georgf , of course!

Here is the latest with navbar links bolded only. Let me know if you feel this will work.

telemetry1after2

[Live site for reference](https://jquinnie.github.io/telemetry-dashboard/probe-dictionary)

georgf:
Ok, that definitely makes it more visible already!
@chutten, does that change make it stand out more to you?

chutten:
It does, but the stats view appears to be broken both in the production site and in the PR one which makes it hard to compare.

chutten:
Ah, `gChannelInfo.any.versions` is empty. `extractChannelInfos` doesn't get around to populating it.

It'll probably work if I select a channel. I'll test it that way.

probe-dictionary/explorer.css
```diff
@@ -122,3 +122,7 @@ body {
 .hidden {
     display: none;
 }
+
+.nav-item {
+    font-weight: bold;
```
chutten, line 127: How did `text-decoration: underline` look? Users on the web are conditioned to look for underlines (links) and boxes (buttons). ...but if it looks odd, we bold is definitely an improvement over the present style.

probe-dictionary/index.html
```diff
@@ -31,11 +31,11 @@
       <button class="navbar-toggler navbar-toggler-right" type="button" data-toggle="collapse" data-target="#navbarCollapse" aria-controls="navbarCollapse" aria-expanded="false" aria-label="Toggle navigation">
         <span class="navbar-toggler-icon"></span>
       </button>
-      <a class="navbar-brand" href="#">Probe dictionary</a>
+      <a class="navbar-brand" href="#">Probe Dictionary</a>
 
       <div class="collapse navbar-collapse" id="navbarCollapse">
-        <ul class="navbar-nav mr-auto">
-          <li class="nav-item active">
+        <ul class="navbar-nav mr-auto navbar-custom">
```
chutten: What is the purpose of `navbar-custom` here? It appears to add no styles or semantic information... but I'm not the most familiar with bootstrap, so maybe it's customary?

probe-dictionary/explorer.css
JQuinnie: @chutten I think I prefer without underline... seems cleaner, take a look at the underline below;

telemetry1after3

Let me know if that's something you want me to include.

probe-dictionary/index.html
JQuinnie: @chutten That was my mistake, a carry over from the last custom style. I have promptly removed.

probe-dictionary/explorer.css
chutten: I agree, the underline doesn't work in this design.

chutten (approved):

georgf:
Thanks for the contribution!

PR #454, opened 2018-03-08
Fix #440: Add button group to dashboard buttons.

### LINKED ISSUE #440: Use button group for top-right dashboard buttons
In #424 Erin suggested a nice visual improvement of using a button group for the top right buttons in the measurements dashboard.
We should do that.

ecomerford:
Hi @georgf, could you review this when you get a chance?

The page is live [here](https://ecomerford.github.io/telemetry-dashboard/). 

I tried to have the permalink button be included in the button group, but it was relegated to the end of the row when I did that. I think because it expands? So I thought better to have it in front as we discussed and remove it from the button group. That's what I've done here. If you have any ideas on keeping it in the front of the row _and_ in the button group, I'm happy to look into it again!

Thank you!

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -22,8 +22,9 @@
     <div class="container-fluid">
         <header>
             <form class="navbar-form inline pull-right">
-                <span class="permalink-control"></span>
                 <div class="btn-group">
+                <a href="#" class="btn btn-default">Get permalink</a>
+                <input type="text" class="input btn btn-default"></span>
                 <a href="https://github.com/mozilla/telemetry-dashboard/issues" class="btn btn-success" title="Report a bug in Telemetry dashboards" target="_blank">Report bug</a>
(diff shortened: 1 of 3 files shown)


georgf:
I like it, except for the separate link button :)
I tried around a little, and this structure seems to be working:
```
<div class="btn-group">
  <a href="#" class="btn btn-default">Get permalink</a>
  <input type="text" class="input btn btn-default">(should be hidden by default)</span>
  <a ...>Report bug</a>
  ...
</div>
```

It will require changing the JavaScript a bit that set's things up. How is that?

georgf:
The button group in the live view looks good.
I think what's missing functionally is:
- hiding the text input box by default
- after the button click, making the short link show up in the text input box

georgf (commented):
As commented above :)

ecomerford:
@georgf I'm going to close this pull request so that I can make commits/see the live results without constantly updating you while I'm still working on this! I'll re-open it once I'm happy with the results.

georgf:
Ah, got it! I wasn't sure if this needed review or not.

PR #465, opened 2018-03-12
Fixed #447: Use checkboxes in dashboards "advanced settings"

### LINKED ISSUE #447: Use checkboxes in dashboards "advanced settings"
The "advanced settings" in the distribution and evolution dashboard should just be checkboxes instead of radio buttons.

This will require a bit of rewording and reordering of elements so that it makes sense.

paucaro:
I have delayed reviewing the code in javascript (until I understood it more) and now I have replaced the radio button with checkboxes in the "advanced settings" dashboard, and here is the live link: https://paulaasto.github.io/telemetry-dashboard/

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -162,29 +162,19 @@
                                     </select> from highest to lowest
                                 </div>
                                 <div>
-                                    <b>Table mode:</b> show histograms as tables of values rather than visual charts:
-                                    <label><input name="table-toggle" type="radio" value="0"> Show histograms</label>
-                                    <label><input name="table-toggle" type="radio" value="1"> Show tables</label>
+                                    <label><input name="table-toggle" type="checkbox" value="0"> <b>Table mode</b></label> Show histograms as tables of values rather than visual charts.
                                 </div>
                                 <div>
-                                    <b>Cumulative mode:</b> show each bucket as the sum of itself and all buckets on its left:
-                                    <label><input name="cumulative-toggle" type="radio" value="0"> Non-cumulative</label>
-                                    <label><input name="cumulative-toggle" type="radio" value="1"> Cumulative</label>
+                                    <label><input name="cumulative-toggle" type="checkbox" value="0"> <b>Cumulative mode</b></label> Show each bucket as the sum of itself and all buckets on its left.
                                 </div>
                                 <div>
-                                    <b>Date range variable:</b> filter submissions by build date range or submission date range:
-                                    <label><input name="build-time-toggle" type="radio" value="0"> Build Date</label>
-                                    <label><input name="build-time-toggle" type="radio" value="1"> Submission Date</label>
+                                    <label><input name="build-time-toggle" type="checkbox" value="0"> <b>Submission date range</b></label> Filter submissions as submission date range rather than build date range.
                                 </div>
                                 <div>
-                                    <b>Sanitize data:</b> hide data points that don't have enough submissions:
-                                    <label><input name="sanitize-toggle" type="radio" value="0"> Don't Sanitize</label>
-                                    <label><input name="sanitize-toggle" type="radio" value="1"> Sanitize</label>
+                                    <label><input name="sanitize-toggle" type="checkbox" value="0"> <b>Sanitize data</b></label> Hide data points that don't have enough submissions.
                                 </div>
                                 <div id="trim-option">
-                                    <b>Trim small buckets:</b> hide buckets on both sides that don't have enough samples:
-                                    <label><input name="trim-toggle" type="radio" value="0"> Don't Trim Buckets</label>
-                                    <label><input name="trim-toggle" type="radio" value="1"> Trim Buckets</label>
+                                    <label><input name="trim-toggle" type="checkbox" value="0"> <b>Trim small buckets</b></label> Hide buckets on both sides that don't have enough samples.
                                 </div>
                                 <span id="date-range-controls" class="form-group" style="width: 230px; display: inline-block; margin: 0">
                                     <span class="input-prepend input-group">
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -55,7 +55,10 @@
                 </h2>
                 <hr></hr>
                 <figure class="col-md-12">
-                  <figcaption id="dist-caption"></figcaption>
+                    <div class="caption-div">
+                        <figcaption id="dist-caption-text" class="overflow"></figcaption>
+                        <figcaption id="dist-caption-link"></figcaption>
+                    </div>
                     <div class="col-md-9" id="plots">
                         <div class="row">
                             <div class="col-md-6" style="text-align: center">
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -159,29 +159,19 @@
                                     </select> from highest to lowest
                                 </div>
                                 <div>
-                                    <b>Table mode:</b> show histograms as tables of values rather than visual charts:
-                                    <label><input name="table-toggle" type="radio" value="0"> Show histograms</label>
-                                    <label><input name="table-toggle" type="radio" value="1"> Show tables</label>
+                                    <label><input name="table-toggle" type="checkbox" value="0"> <b>Table mode</b></label> Show histograms as tables of values rather than visual charts.
                                 </div>
                                 <div>
-                                    <b>Cumulative mode:</b> show each bucket as the sum of itself and all buckets on its left:
-                                    <label><input name="cumulative-toggle" type="radio" value="0"> Non-cumulative</label>
-                                    <label><input name="cumulative-toggle" type="radio" value="1"> Cumulative</label>
+                                    <label><input name="cumulative-toggle" type="checkbox" value="0"> <b>Cumulative mode</b></label> Show each bucket as the sum of itself and all buckets on its left.
                                 </div>
                                 <div>
-                                    <b>Date range variable:</b> filter submissions by build date range or submission date range:
-                                    <label><input name="build-time-toggle" type="radio" value="0"> Build Date</label>
-                                    <label><input name="build-time-toggle" type="radio" value="1"> Submission Date</label>
+                                    <label><input name="build-time-toggle" type="checkbox" value="0"> <b>Submission date range</b></label> Filter submissions as submission date range rather than build date range.
                                 </div>
                                 <div>
-                                    <b>Sanitize data:</b> hide data points that don't have enough submissions:
-                                    <label><input name="sanitize-toggle" type="radio" value="0"> Don't Sanitize</label>
-                                    <label><input name="sanitize-toggle" type="radio" value="1"> Sanitize</label>
+                                    <label><input name="sanitize-toggle" type="checkbox" value="0"> <b>Sanitize data</b></label> Hide data points that don't have enough submissions.
                                 </div>
                                 <div id="trim-option">
-                                    <b>Trim small buckets:</b> hide buckets on both sides that don't have enough samples:
-                                    <label><input name="trim-toggle" type="radio" value="0"> Don't Trim Buckets</label>
-                                    <label><input name="trim-toggle" type="radio" value="1"> Trim Buckets</label>
+                                    <label><input name="trim-toggle" type="checkbox" value="0"> <b>Trim small buckets</b></label> Hide buckets on both sides that don't have enough samples.
                                 </div>
                                 <span id="date-range-controls" class="form-group" style="width: 230px; display: inline-block; margin: 0">
                                     <span class="input-prepend input-group">
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -162,19 +162,24 @@
                                     </select> from highest to lowest
                                 </div>
                                 <div>
-                                    <label><input name="table-toggle" type="checkbox" value="0"> <b>Table mode</b></label> Show histograms as tables of values rather than visual charts.
+                                    <label><input name="table-toggle" type="checkbox"> <b>Table mode</b> </label>
+                                    Show histograms as tables of values rather than visual charts.
                                 </div>
                                 <div>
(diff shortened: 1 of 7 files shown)


new-pipeline/dist.html
```diff
@@ -159,29 +159,19 @@ <h3><select id="selected-key4" class="multiselect" title="Selected key"></select
                                     </select> from highest to lowest
                                 </div>
                                 <div>
-                                    <b>Table mode:</b> show histograms as tables of values rather than visual charts:
-                                    <label><input name="table-toggle" type="radio" value="0"> Show histograms</label>
-                                    <label><input name="table-toggle" type="radio" value="1"> Show tables</label>
+                                    <label><input name="table-toggle" type="checkbox" value="0"> <b>Table mode</b></label> Show histograms as tables of values rather than visual charts.
```
georgf: This line is a little long now. Can we put the description on a new line?
Similarly for the elements below.

new-pipeline/src/evo.js
```diff
@@ -811,9 +811,16 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
     .click(function () {
       var newUseSubmissionDate = $("input[name=build-time-toggle]:checked")
         .val() !== "0" ? 0 : 1;
-      $("input[name=build-time-toggle][value=" + newUseSubmissionDate + "]")
-        .prop("checked", true)
-        .trigger("change");
+      if(newUseSubmissionDate == 0){
```
georgf: A minor thing: We use spaces around braces, e.g. `if (...) {`.

georgf (commented):
This looks good, thanks!

From testing it is working fine. @chutten is more familiar with this dashboard though, so i'm asking him to double-check this.

I only have two small style issues below.

new-pipeline/dist.html
```diff
@@ -159,29 +159,19 @@ <h3><select id="selected-key4" class="multiselect" title="Selected key"></select
...
-                                    <b>Date range variable:</b> filter submissions by build date range or submission date range:
-                                    <label><input name="build-time-toggle" type="radio" value="0"> Build Date</label>
-                                    <label><input name="build-time-toggle" type="radio" value="1"> Submission Date</label>
+                                    <label><input name="build-time-toggle" type="checkbox" value="0"> <b>Submission date range</b></label> Filter submissions as submission date range rather than build date range.
                                 </div>
                                 <div>
-                                    <b>Sanitize data:</b> hide data points that don't have enough submissions:
-                                    <label><input name="sanitize-toggle" type="radio" value="0"> Don't Sanitize</label>
-                                    <label><input name="sanitize-toggle" type="radio" value="1"> Sanitize</label>
+                                    <label><input name="sanitize-toggle" type="checkbox" value="0"> <b>Sanitize data</b></label> Hide data points that don't have enough submissions.
```
chutten: The default is to sanitize, so this checkbox should be checked.

new-pipeline/dist.html
```diff
@@ -159,29 +159,19 @@ <h3><select id="selected-key4" class="multiselect" title="Selected key"></select
...
                                 </div>
                                 <div id="trim-option">
-                                    <b>Trim small buckets:</b> hide buckets on both sides that don't have enough samples:
-                                    <label><input name="trim-toggle" type="radio" value="0"> Don't Trim Buckets</label>
-                                    <label><input name="trim-toggle" type="radio" value="1"> Trim Buckets</label>
+                                    <label><input name="trim-toggle" type="checkbox" value="0"> <b>Trim small buckets</b></label> Hide buckets on both sides that don't have enough samples.
```
chutten: The default is to trim, so this checkbox should be checked.

new-pipeline/evo.html
```diff
@@ -79,14 +79,10 @@ <h2 id="sample-counts-title" style="font-size: 18px; margin-bottom: 0"></h2>
...
-                                    Evolution variable: show evolution over build date or submission date:
-                                    <label><input name="build-time-toggle" type="radio" value="0"> Build Date</label>
-                                    <label><input name="build-time-toggle" type="radio" value="1"> Submission Date</label>
+                                    <label><input name="build-time-toggle" type="checkbox" value="0"> <b>Build date range</b></label> Filter submissions as build date range rather than submission date range.
                                 </div>
                                 <div>
-                                    Sanitize data: hide data points that don't have enough submissions:
-                                    <label><input name="sanitize-toggle" type="radio" value="0"> Don't Sanitize</label>
-                                    <label><input name="sanitize-toggle" type="radio" value="1"> Sanitize</label>
+                                    <label><input name="sanitize-toggle" type="checkbox" value="0"> <b>Sanitize data</b></label> Hide data points that don't have enough submissions.
```
chutten: The default is to sanitize, so this checkbox should be checked.

new-pipeline/dist.html
```diff
@@ -159,29 +159,19 @@ <h3><select id="selected-key4" class="multiselect" title="Selected key"></select
...
-                                    <b>Table mode:</b> show histograms as tables of values rather than visual charts:
-                                    <label><input name="table-toggle" type="radio" value="0"> Show histograms</label>
-                                    <label><input name="table-toggle" type="radio" value="1"> Show tables</label>
+                                    <label><input name="table-toggle" type="checkbox" value="0"> <b>Table mode</b></label> Show histograms as tables of values rather than visual charts.
                                 </div>
                                 <div>
-                                    <b>Cumulative mode:</b> show each bucket as the sum of itself and all buckets on its left:
-                                    <label><input name="cumulative-toggle" type="radio" value="0"> Non-cumulative</label>
-                                    <label><input name="cumulative-toggle" type="radio" value="1"> Cumulative</label>
+                                    <label><input name="cumulative-toggle" type="checkbox" value="0"> <b>Cumulative mode</b></label> Show each bucket as the sum of itself and all buckets on its left.
```
chutten: The standards-compliant way to check a checkbox is to provide the boolean attribute `checked`. If you want all of these to initialize as unchecked (the radio buttons did start out blank, so that'd be fine) then you can omit the value altogether.

new-pipeline/src/dist.js
```diff
@@ -67,11 +67,11 @@ $(function () {
       .multiselect("select", gInitialPageState.sort_keys);
     $("input[name=table-toggle][value=" + (gInitialPageState.table !==
```
chutten: `gInitialPageState` contains the configuration of how the user previously looked at the page. They may have changed all of the advanced settings in any number of ways.

Now that we're using checkboxes instead of radio buttons, this will have to be reworked. It should be more straightforward now as we simply need to set a checkbox to `checked` if the page state says it's non-zero, or uncheck it otherwise.

So something like

```js
$("input[name=cumulative-toggle]").checked = gInitialPageState.cumulative !== 0;
$("input[name=cumulative-toggle]").trigger("change");
```

for all of the checkboxes. I'm not sure we need to trigger a `change` on all of them, but we'd better leave it in there for now, just in case.

new-pipeline/src/evo.js
```diff
@@ -1021,7 +1028,7 @@ function saveStateToUrlAndCookie() {
 
   // If advanced settings are not at their defaults, display a notice in the panel header
   if (gInitialPageState.use_submission_date !== 0 || gInitialPageState.sanitize !==
```
chutten: Default sanitize setting is to sanitize, so this should probably be left more or less the same.

new-pipeline/src/dist.js
```diff
@@ -1850,8 +1850,8 @@ function saveStateToUrlAndCookie() {
   }
 
   var isKeyedHistogram = $("#sort-keys-option").css("display") !== "none";
-  if (gInitialPageState.use_submission_date !== 0 || gInitialPageState.table !==
-    0 || gInitialPageState.cumulative !== 0 || gInitialPageState.trim !== 1 ||
+  if (gInitialPageState.use_submission_date == 0 || gInitialPageState.table ==
```
chutten: The current defaults should be left alone in order to not surprise users who are used to the default behaviour.

chutten (commented):
A fine start.

If we can rework this to use `checked` instead of `val()` I think we'll be getting closer.

paucaro:
Thanks @chutten and @georgf for the revision, I have made the corresponding changes and now I understand the code better than before. I would be very grateful if you reviewed the changes I made.

new-pipeline/src/dist.js
```diff
@@ -65,25 +65,20 @@ $(function () {
     // Initialize setting values from the page state
     $("#sort-keys")
       .multiselect("select", gInitialPageState.sort_keys);
-    $("input[name=table-toggle][value=" + (gInitialPageState.table !==
-        0 ? 1 : 0) + "]")
-      .prop("checked", true)
+    $("input[name=table-toggle]")
+      .prop("checked", gInitialPageState.table !== 0 ? true : false)
```
chutten: The result of `gInitialPageState.table !== 0` is a boolean, so you don't need this ternary expression (or the other similar ones) to turn it to true/false.

new-pipeline/src/dist.js
chutten: These conditions should probably be left alone. I think this is why I'm seeing, on an unmodified load, the Advanced Settings saying "(modified)" even though I haven't modified them.

new-pipeline/src/evo.js
```diff
@@ -505,10 +504,9 @@ function calculateEvolutions(callback) {
   channelVersions.forEach(function (channelVersion) {
     var parts = channelVersion.split("/"); //wip: fix this
     getHistogramEvolutionLines(parts[0], parts[1], measure, aggregates,
-      filterSets, $("input[name=sanitize-toggle]:checked")
-      .val() !== "0", $("input[name=build-time-toggle]:checked")
-      .val() !== "0",
-      function (newLinesMap, newDescription) {
+      filterSets, $("input[name=sanitize-toggle]").is(":checked")
+      , $("input[name=build-time-toggle]").is(":checked")
```
chutten: The style in this file is to put the commas at the ends of the previous line, not at the beginning of the next line.

new-pipeline/src/evo.js
```diff
@@ -809,10 +807,10 @@ function displayEvolution(target, lines, usePercentages, plotOptions) {
...
     .attr("text-decoration", "underline")
     .click(function () {
-      var newUseSubmissionDate = $("input[name=build-time-toggle]:checked")
-        .val() !== "0" ? 0 : 1;
-      $("input[name=build-time-toggle][value=" + newUseSubmissionDate + "]")
-        .prop("checked", true)
+      var newUseSubmissionDate = $("input[name=build-time-toggle]")
+        .is(":checked") ? 0 : 1;
+      $("input[name=build-time-toggle]")
+        .prop("checked", newUseSubmissionDate != 0 ? true : false)
```
chutten: Now that we're using `checked` directly, I think we can simplify this toggling behaviour to look something like:

```js
let buildTimeToggle = $("input[name=build-time-toggle]");
buildTimeToggle.prop("checked", !buildTimeToggle.prop("checked"));
```

new-pipeline/src/evo.js
```diff
@@ -1021,7 +1019,7 @@ function saveStateToUrlAndCookie() {
 
   // If advanced settings are not at their defaults, display a notice in the panel header
   if (gInitialPageState.use_submission_date !== 0 || gInitialPageState.sanitize !==
-    1) {
+    0) {
```
chutten: I think this needs to remain unchanged

new-pipeline/style/dashboards.css
```diff
@@ -123,10 +123,26 @@
...
+  text-overflow: ellipsis;
   font-size: 1.5em;
   text-align: center;
-  margin: 0.1em;
+}
+
+#dist-caption-link, #evo-caption-link {
+  color : black;
+  font-size: 1.5em;
+  display: inline-block;
```
chutten, line 145: I think the changes in this file are unrelated?

new-pipeline/tutorial.html
```diff
@@ -13,21 +13,80 @@
     <link rel="stylesheet" href="style/jquery.tocify.css">
     <style>
     #tocify-header0 > .tocify-item { font-weight: bold; } /* bold the top-level entries */
+
```
chutten, line 16: And the changes to this file seem unrelated as well.

chutten (changes requested):
Looks good!

Seems as though there are a few unrelated commits that may have snuck into this PR uninvited. The relevant changes are good, though. We're nearly there.

new-pipeline/style/dashboards.css
paucaro: Hi @chutten , you are rigth; these changes are not related to the task, apparently when I did pull in my fork repository they were included in the commit as well.

paucaro:
Thanks for the recommendations made @chutten , I made the required changes, if there was any other problem, issue or recommendation, I would be happy to do it.

chutten (changes requested):
It behaves very well in my testing, and the JS and HTML changes look good.

There are still unrelated changes in this pull request (to dashboard.css and tutorial.html, and to do with the caption-div stuff) that will have to be pulled out before I can merge this in, though.

paucaro:
@chutten @georgf I have reviewed why these changes appear in the files: dashborad.js dashborad.css and tutorial.html, and I have realized that they are not really my changes, they are changes made by other tasks as you can see in the image. So I do not know what I could do about it. I would be very grateful if you gave me some advice or recommendations about it.
Thank you.
image

chutten:
One thing you could try is to copy your current branch state to a new branch, and then use an interactive rebase to pick and chose which commits you want.

`git checkout -b newbranch` will create a new branch `newbranch` and switch you over to it. All of your old commits are on both the old and the new branch.

`git rebase -i HEAD~7` this will allow you to interactively rebase the top 7 commits on your current branch. You can remove, reorder, squash... whatever you want until it looks exactly the way you want it to.

Then you (force-)push that edited branch to the remote branch of your choice.

Does this make sense?

A nice effect of this approach is that your old branch with all the commits the way they are today has been left totally alone. So if anything goes wrong, you can get back to where you were without diving into reflog.

chutten:
We're taking this over to #475

PR #466, opened 2018-03-13
Fix #440: Add button group to top-right dashboard buttons

### LINKED ISSUE #440: Use button group for top-right dashboard buttons

ecomerford:
Hi @georgf and/or @chutten, could you review these changes when you have a chance? I believe this is what you had in mind, Georg!

You can view the changes live [here](https://ecomerford.github.io/telemetry-dashboard/).

Thanks! :)

### DIFF
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -127,7 +127,7 @@ $(document)
     }
 
     // Permalink control
-    $(".permalink-control input")
+    $(".permalink-control #hidden")
       .hide()
       .focus(function () {
         // Workaround for broken selection: http://stackoverflow.com/questions/5797539
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -163,12 +163,12 @@ $(document)
           async:false
         });
         document.execCommand('copy');
+      });
         // Work around to force the clipboard to hold short URL from permalink button.
         document.addEventListener("copy", e => {
           e.clipboardData.setData("text/plain", shortUrl);
           e.preventDefault();
         });
-      });
      });
 
 // Load the current state from the URL, or the cookie if the URL is not specified
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -127,8 +127,17 @@ $(document)
     }
 
     // Permalink control
-    $(".permalink-control input")
+    $(".permalink-control.input")
       .hide()
+   .focus(function () {
+        // Workaround for broken selection: http://stackoverflow.com/questions/5797539
+        var $this = $(this);
+        $this.select()
+          .mouseup(function () {
+            $this.unbind("mouseup");
+            return false;
+          });
+     });
     $(".permalink-control")
       .click(function () {
         var $this = $(this);
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -129,7 +129,7 @@ $(document)
     // Permalink control
     $(".permalink-control.input")
       .hide()
-   .focus(function () {
+      .focus(function () {
         // Workaround for broken selection: http://stackoverflow.com/questions/5797539
         var $this = $(this);
         $this.select()
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -138,7 +138,7 @@ $(document)
             return false;
           });
       });
-    $(".permalink-control btn")
+    $(".permalink-control")
       .click(function () {
         var $this = $(this);
         $.ajax({
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -130,14 +130,14 @@ $(document)
     $(".permalink-text")
       .hide()
       .focus(function () {
-       // Workaround for broken selection: http://stackoverflow.com/questions/5797539
-       var $this = $(this);
-       $this.select()
-         .mouseup(function () {
-           $this.unbind("mouseup");
-           return false;
+        // Workaround for broken selection: http://stackoverflow.com/questions/5797539
+        var $this = $(this);
+        $this.select()
+        .mouseup(function () {
+          $this.unbind("mouseup");
+          return false;
          });
-     });
+      });
     $(".permalink-button")
       .click(function () {
         var $this = $(this);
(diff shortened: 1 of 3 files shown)


new-pipeline/src/dashboards.js
```diff
@@ -127,25 +127,18 @@ $(document)
...
-        '<div class="input-group">' +
-        '    <span class="input-group-btn"><button type="button" class="btn btn-default" title="Get Permalink"><i class="fa fa-link"></i> Get Permalink</button></span>' +
-        '    <input type="text" class="form-control">' +
-        '</div>'
-      );
-    $(".permalink-control input")
+    $(".permalink-control.input")
       .hide()
-      .focus(function () {
+   .focus(function () {
```
chutten: Should be indented to `.hide()`'s level

new-pipeline/src/dashboards.js
```diff
@@ -127,25 +127,18 @@ $(document)
...
         // Workaround for broken selection: http://stackoverflow.com/questions/5797539
         var $this = $(this);
         $this.select()
           .mouseup(function () {
             $this.unbind("mouseup");
             return false;
           });
-      });
-    $(".permalink-control button")
+     });
```
chutten: Indented one space too far.

chutten (changes requested):
Somehow this change breaks the auto-copy-to-clipboard code. In the current dashboard, clicking on the permalink will copy the returned shortlink into the clipboard for immediate paste.

This has always been a tricky thing to get working, since the `document.execCommand` has to come after the shortlink is returned, but still during a user gesture handler (in this case the `click`)

Maybe it is because this is now on `.permalink-control` (which applies to multiple elements now, both the `a` and the `input`) instead of `.permalink-control.btn`which actually services the user gesture?

chutten:
Oh, and I missed this in my review, but `.permalink-control` is now less useful of a class. It might make more sense to invent `.permalink-button` and `.permalink-text` (or similar) classes to uniquely identify the two parts.

chutten:
Ack, and another thing I missed until I was playing with it just now (Sorry!) was that clicking on the "Get Permalink" button now seems to navigate the page to `https://ecomerford.github.io/telemetry-dashboard/new-pipeline/dist.html#` from what it used to be (in my case, `https://ecomerford.github.io/telemetry-dashboard/new-pipeline/dist.html#!cumulative=0&end_date=2018-03-08&keys=__none__!__none__!__none__&max_channel_version=nightly%252F60&measure=GC_MS&min_channel_version=null&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=2018-01-22&table=0&trim=1&use_submission_date=0`). 

I'm not sure what's going on there, but it's certainly interesting!

ecomerford:
Thanks, @chutten! I think you're right re: making new classes. I'll work on that and getting the auto-copy function back up and running. 

I didn't notice the change in URL but I'll try to investigate. Thanks for pointing it out!

ecomerford:
Hm, the auto-copy seems to be working for me. I'll make the changes to the classes to see if that irons everything out and then check back with you.

ecomerford:
@chutten I've updated the classes and the class selectors. The auto-copy works for me, though it only works if you click it once the entire page loads. I'm not sure if this is a glitch on my end or not...

ecomerford:
@chutten actually, the auto-copy has only worked a few times, not worked other times. I'm not sure why. I'll keep working!

ecomerford:
@chutten Okay, it seems as though the auto-copy is working on Chrome and Edge browsers, but not on Firefox. You mentioned that it might have to do with both classes being selected for the `execCommand` call, but now that I've changed that, the problem persists.

I can't figure out why I'm having this problem. Do you or @georgf have any ideas?

Thank you!

chutten:
This is very strange. `document.queryCommandSupported('copy')` comes back `true`, and during the handler `document.queryCommandEnabled('copy')` comes back `true`.

Even `document.execCommand('copy')` returns `true` (success).

And yet nothing enters the system clipboard....

Well, let's try a workaround. When there's a copy, even one that's empty, we get a ["copy" event](https://developer.mozilla.org/en-US/docs/Web/Events/copy) on `document`. So we can tie in like

```js
document.addEventListener("copy", e => {
  e.clipboardData.setData("text/plain", <the shortlink from the input>);
  e.preventDefault();
});
```

Maybe that will take care of things.

While you're back in the code, make sure it's at two-space indentation levels. From the patch it looks as though some of it might have acquired an extra space here and there (check the "Files changed" tab at the top of this page)

ecomerford:
It doesn't seem to be working for me. Did I use the correct variable (`shortUrl`)?

georgf:
Visually this looks good now!
I'll leave the technical feedback here to @chutten.

new-pipeline/src/dashboards.js
```diff
@@ -163,8 +163,14 @@ $(document)
           async:false
         });
         document.execCommand('copy');
+        });
       });
-   });
+
+// Work around to force the clipboard to hold short URL from permalink button.
+document.addEventListener("copy", e => {
+  e.clipboardData.setData("text/plain", shortUrl);
```
chutten: `shortUrl` won't be present in this context, since it only lives about as long as the `success` handler from line 152 is running. 

Since the success handler sets the value of the `input` to `shortUrl`, you should be able to read it from there at the point in time the copy event listener runs.

Make sense?

ecomerford: Yes, thank you!

ecomerford: By the way, do the extra indents seem to be corrected now?

chutten: The indents seem better. There are two little indents that need fixing, though.

Lines 136-139 should be two spaces further in (`.mouseup` up to the closing `)};`)
And Lines 166 and 167 should be two spaces further back (the two `)};` should be indented on the original level, not inline with the method bodies.

So it should look like
```js
this.doSomething()
  .andSomethingElse(function () {
    theFunctionBodyGoesHere();
  }); //should line up with "." instead of with "the"
```

ecomerford:
@chutten I've tried it in a few different spots now -- in and outside of the `success` function -- but neither make the auto-copy work on Firefox.

ecomerford:
I noticed that in the current version of the dashboard, the button generates a short-link like this: "https://mzl.la/2plR8Z0", whereas in my version, the short-link always appears as a bit.ly link, i.e. "https://bit.ly/2I3R4Ff". Could this be a clue?

chutten:
Instead of using `shortUrl` (which may complete its lifetime before the event handler runs) maybe  `$(".permalink-text").val()` would work, since it'll have been updated by then?

As for the mzl.la vs bit.ly difference, that's because bit.ly shortens things to mzl.la if they end in mozilla.org or mozilla.com for us. (not sure what the arrangement is there, but it means I know before clicking it when a shortlink points to a mozilla address)

ecomerford:
Thanks, @chutten . Going to work on this a bit more and open up a new PR.

PR #467, opened 2018-03-13
Fix #409: Show the distribution and evolution views as tabs in dashboard

### LINKED ISSUE #409: Show the distribution and evolution views as tabs in dashboard

Insiyaa:
Please review :)

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -23,13 +23,16 @@
         <header>
             <form class="navbar-form inline pull-right">
                 <a href="https://github.com/mozilla/telemetry-dashboard/issues" class="btn btn-success" title="Report a bug in Telemetry dashboards" target="_blank">Report bug</a>
-                <a href="./evo.html" class="btn btn-default" id="switch-views" title="Switch to the evolution view">Evolution View</a>
                 <a href="/" class="btn btn-default btn-primary" title="Landing page for Telemetry dashboards">Dashboards Home</a>
                 <a href="./tutorial.html#HistogramDashboard" class="btn btn-default" id="tutorial" title="Usage tutorial for Telemetry dashboards">Usage Tutorial</a>
                 <span class="permalink-control"></span>
             </form>
             <h1>Measurement Dashboard</h1>
         </header>
+        <ul class="nav nav-tabs">
+            <li class="active" ><a href="#">Distribution</a></li>
+            <li><a href="./evo.html">Evolution</a></li>
+        </ul>
         <div class="row">
             <div class="col-md-12">
                 <h2 style="font-size: 18px; margin-bottom: 0">
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -16,6 +16,11 @@
     <link rel="stylesheet" type="text/css" href="style/metricsgraphics.css" />
 
     <link rel="stylesheet" type="text/css" href="style/dashboards.css" />
+    <style>
+        .nav-tabs a {
+            color: black;
+        }
+    </style>
     <script async src="https://www.googletagmanager.com/gtag/js?id=UA-49796218-2"></script>
 </head>
 <body>
@@ -30,8 +35,8 @@
             <h1>Measurement Dashboard</h1>
         </header>
         <ul class="nav nav-tabs">
-            <li class="active" ><a href="#">Distribution</a></li>
-            <li><a href="./evo.html">Evolution</a></li>
+            <li class="active" ><a href="#"><i class="fa fa-bar-chart"></i><strong>  Distribution</strong></a></li>
+            <li><a href="./evo.html"><i class="fa fa-line-chart"></i><strong>  Evolution</strong></a></li>
         </ul>
         <div class="row">
             <div class="col-md-12">
--- a/new-pipeline/tutorial.html
+++ b/new-pipeline/tutorial.html
@@ -13,21 +13,61 @@
     <link rel="stylesheet" href="style/jquery.tocify.css">
     <style>
     #tocify-header0 > .tocify-item { font-weight: bold; } /* bold the top-level entries */
+    body {
+        padding: 20px 5px 5px 5px;
+    }
+    .active {
+        background: rgb(241, 236, 236);
+        border-left: 5px solid #286090;
+    }
+    hr {
+        margin: 0 auto;
+    }
+    nav a{
+        color:black;
+    }
+    nav .nav-stacked li a{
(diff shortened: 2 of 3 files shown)


Insiyaa:
Link: https://insiyaa.github.io/telemetry-dashboard/

new-pipeline/dist.html
```diff
@@ -23,185 +23,266 @@
         <header>
             <form class="navbar-form inline pull-right">
                 <a href="https://github.com/mozilla/telemetry-dashboard/issues" class="btn btn-success" title="Report a bug in Telemetry dashboards" target="_blank">Report bug</a>
-                <a href="./evo.html" class="btn btn-default" id="switch-views" title="Switch to the evolution view">Evolution View</a>
+                <!-- <a href="./evo.html" class="btn btn-default" id="switch-views" title="Switch to the evolution view">Evolution View</a> -->
```
chutten: Leaving commented-out code in the tree is unnecessary (we make one exception for warning notices, so we know where to put them in a hurry)

new-pipeline/tutorial.html
```diff
@@ -13,21 +13,80 @@
     <link rel="stylesheet" href="style/jquery.tocify.css">
```
chutten, line 13: Are changes to the tutorial necessary?

new-pipeline/dist.html
```diff
@@ -220,6 +301,7 @@ <h3><select id="selected-key4" class="multiselect" title="Selected key"></select
     <script src="../v2/telemetry.js"></script>
     <script src="src/dashboards.js"></script>
     <script src="src/dist.js"></script>
+    <script src="src/evo.js"></script>
```
chutten: These scripts interfere with each other by overriding each others' global storage. I'm not sure dist.js and evo.js can be present in the same context in the way they are written today.

new-pipeline/dist.html
```diff
@@ -23,185 +23,266 @@
...
-                                        <input type="text" name="date-range" style="width: 100%" id="date-range" class="form-control date-range">
-                                    </span>
-                                </span>
-                                <span id="range-bar" class="pull-right"></span>
-                            </form>
+                            </div>
                         </div>
                     </div>
                 </div>
+                <div class="busy-indicator">
```
chutten: The busy indicator needs to be in the correct stacking context to be displayed without padding on the left. I don't think it can be moved in here.

new-pipeline/dist.html
```diff
@@ -23,185 +23,266 @@
...
+            <li class="active" ><a href="#dist" data-toggle="tab"><strong>Distribution</strong></a></li>
+            <li><a data-toggle="tab" href="#evo"><strong>Evolution</strong></a></li>
+        </ul>
+        <div class="tab-content">
+            <div id="dist" class="tab-pane fade in active">
+                <div class="row">
+                    <div class="col-md-12">
+                        <h2 style="font-size: 18px; margin-bottom: 0">
+                            <span style="font-weight: bold">
+                                <select id="measure" class="multiselect" title="Value being measured"></select>
```
chutten: `measure` is a duplicated id. ids must be unique in the document, otherwise the first one wins.

There are other duplicated ids.

chutten (changes requested):
Does this solution work on your fork? In trying to load it from the link provided, it wouldn't initialize, and some UI elements were misplaced.

chutten:
Upon closer inspection of #409, we don't need to merge dist and evo together to fix the bug. We just need to add a tab control to both to make them _look_ as though they're the same page. Clicking the tabs will still change location from dist.html to evo.html (and back again) the way that the current buttons at the top do. It's just that instead of buttons in the top we'll have tabs in the middle.

Does this make sense @Insiyaa ?

Insiyaa:
I got it. Sorry, I misinterpreted the issue. I'll make the required changes.

chutten:
No, no, it's our fault for not describing the problem clearly. To us, who have lived in this code quite a bit, we wouldn't have thought to try and bring the two together, so it didn't occur to us to mention it. An oversight on our part.

Insiyaa:
Please review. I removed the top buttons to switch between views.
Link: https://insiyaa.github.io/telemetry-dashboard/

georgf:
I only took a quick link at the live link.
Visually that looks nice, cheers!

There are smaller visual things i noticed:
- [The current dashboard](https://telemetry.mozilla.org/new-pipeline/dist.html) now uses font-awesome icons for the distribution & evolution buttons, can you add them to the tabs here?
- The text color for the inactive tab is blue (link styling), can we make it black too?

Insiyaa:
Yeah done! Please review :)

new-pipeline/dist.html
```diff
@@ -16,20 +16,28 @@
     <link rel="stylesheet" type="text/css" href="style/metricsgraphics.css" />
 
     <link rel="stylesheet" type="text/css" href="style/dashboards.css" />
+    <style>
```
chutten, line 19: This style should be within a current stylesheet, maybe in dashboard.css

chutten (changes requested):
There are a couple of changes needed, but I like the look already.

1) You should rebase your code atop the latest code from `mozilla/gh-pages`. This will ensure we'll be able to merge your code when it's done, and will make comparing your live instance to the real live instance much easier.

2) The active tab label and icon are grey and the inactive ones are black. This is the reverse of what I'd expect since the rest of the active content is black on the page (like the multiselects and stuff) and "disabled" states are usually displayed with grey. Can we make it so that the active tab is black and the inactive tab is grey?

3) putting the shared styles in the shared stylesheet as I mentioned inline

4) There are changes to tutorial.html in this pull request that I don't think are related to the design change? They'll need to be removed before this PR can be merged.

5) This pushes the content down somewhat. In order to make sure the plot (the important content) appears "above the fold" (ie, without scrolling) we may need to take out some margins somewhere. Once you rebase you'll probably find some space we can take from the `margin-bottom` of `div.caption-div`

Does this make sense?

new-pipeline/tutorial.html
Insiyaa: These are from earlier PR, I can't understand why they are still showing up...

Insiyaa:
https://insiyaa.github.io/telemetry-dashboard/new-pipeline/dist.html 
I was stuck with rebasing so deleted the repo and forked again... Commits aren't showing up?

chutten:
It now says this PR is from an "unknown repository". Deleting and reforking probably caused Github to become confused. You might have to close this PR and open a new one from your new repo.

Issue #456, opened 2018-03-08
Add a clear top bar to Telemetry Portal


shreya99oak:
@georgf  I'll start working on this!
To add-
 A suitable colored top bar which will include the title "Telemetry portal" and "report bug" on the right.

kdrosado:
@georgf , @robhudson , Has #468 been assigned? If not, can I work on it?

georgf:
Hi, yes, it has been assigned :)
You can tell by either the "assigned" label on the right or if a person is listed under "assignees".

There are more [available issues](https://github.com/mozilla/telemetry-dashboard/labels/mentored). If you can't find any, just let us know!

gargshruti30:
@georgf Can I work on this issue ?

wlach:
I think this was fixed at some point.

PR #468, opened 2018-03-13
Added a clear top bar for telemetry portal

shreya99oak:
#456   Pls. review the changes..Thanks!


shreya99oak:
Live link for the same....https://shreya99oak.github.io/telemetry-dashboard-probe-dictionary/

georgf (commented):
I see that changes to the probe-dictionary are in this, can you remove them from this branch? :)

georgf (commented):
On the visual style, this looks like a good direction!

Two asks here:
- Can you increase the margin between the content and the heading a little bit? It looks a little squashed.
- Can you use a lighter gray? Ideally this is only a light color different from the content background, so the user is more guided to the content.

shreya99oak:
Sure I'll make the modifications:)

shreya99oak:
I saw the 'changed files', the probe dictionary changes don't appear....the live link shows changes for both the issues though

shreya99oak:
How does this look?


new

probe-dictionary/explorer.css
```diff
@@ -122,3 +122,15 @@ body {
 .hidden {
     display: none;
 }
+
+.contrast:after {
```
georgf: I think this change should be in `new-pipeline/style/dashboards.css`, so we don't need to add a new CSS file to `index.html`.
We just need a class name here, i.e. `.contract {`. 
`.contrast` is a little bit generic though and it is not clear from the name what this is for. How about e.g. `portal-header-bar`?

probe-dictionary/explorer.css
```diff
@@ -122,3 +122,15 @@ body {
...
+    content: "";
```
georgf: Do we need this change?
We should probably be able to just use `background-color`, `margin` and `padding` on this element.
Let's see how that works, we might need to adjust the surrounding elements a little too?

probe-dictionary/index.html
```diff
@@ -167,7 +167,7 @@ <h2 id="detail-probe-name">Probe name</h2>
         <tr><td class="fit pr-2">High:</td><td id="detail-histogram-high" class="grow"></td></tr>
         <tr><td class="fit pr-2">Methods:</td><td id="detail-event-methods" class="grow"></td></tr>
         <tr><td class="fit pr-2">Objects:</td><td id="detail-event-objects" class="grow"></td></tr>
-        <tr><td class="fit pr-2">Extra keys:</td><td id="detail-event-extra-keys" class="grow"></td></tr>
+        <tr><td class="fit pr-2">Extra keys:</td><td id="detail-event-extra-keys" class="grow"></td></tr>
```
georgf: I think this change is not needed here in this pull request.

georgf (commented):
Visually i think this works.
I have some comments on the implementation below, let's sort that out :)

probe-dictionary/explorer.css
shreya99oak: @georgf Ya actually I had tried that, but a default margin was set up on the page ,so I used html { margin :0} but the background color didn't override the margin. Hence I wrote this...I'll still try using margin and padding in the code!

shreya99oak:
Will make the remaining changes!

shreya99oak:
@georgf @chutten , I have made the changes. Pls. review!

chutten:
@georgf is on vacation, but I can take a look.

Can you put this up as a live instance? I tried the previous link and it was showing an older version of the code.

index.html
```diff
@@ -9,17 +9,16 @@
...
 </head>
 <body>
     <div class="container-fluid">
-        <header>
+        <header class="portal-header-bar">
             <form class="navbar-form inline pull-right">
                 <a href="https://github.com/mozilla/telemetry-dashboard/issues" class="btn" title="Report bugs in Telemetry dashboards">Report bug</a>
             </form>
-            <h2>Telemetry portal</h2>
+            <h2>Telemetry portal</h2><br>
```
chutten: What purpose is this `br` serving here? Could it be replaced with CSS?

new-pipeline/style/dashboards.css
```diff
@@ -151,3 +151,10 @@
 .measure-headers th {
   text-align: right;
 }
+
+.portal-header-bar {
+  background-color: #EBEBEB;
+  margin-left: -15px;
+  margin-right: -15px;  
+  padding: 1px 15px 1px;
```
chutten: I'm confused about this use of padding and margin. Could you explain to me its purpose?

chutten (commented):
Please rebase your changes against the latest `mozilla/gh-pages` content.

To do this try adding `moz` as a remote using `git remote add`. Then you can use something like `git rebase --onto` or `git branch --set-upstream-to=` to enable you to push code that will merge cleanly.

index.html
shreya99oak: The line break was to increase the spacing between the content and heading, as it was looking a bit squashed

new-pipeline/style/dashboards.css
shreya99oak: @chutten  The page has a default margin, so to override that I had to use margin and padding...hence margin gets a negative value, the padding was to increase the width as it looked a bit compact...Earlier I had used " :after setting content: " " and z-index=-1..but Georg had asked me to use margin and padding hence I made the above changes

shreya99oak: The branch now tracks moz/gh-pages....I'm not able to provide a live instance as somehow the github page hasn't updated,what could be the reason?

index.html
shreya99oak: Removed <br> It works without that!

shreya99oak:
@chutten Pls. review this:)... here is a live instance https://shreya99oak.github.io/telemetry-dashboard-probe/

chutten (commented):
Is it just me, or is the "Report a bug" link now further up than it used to be? It should be roughly inline with the header (either the centre lines should align, or the baselines)

shreya99oak:
Oh right,had missed that! I'll work on it

shreya99oak:
@chutten Made the changes, it now is line with the base of telemetry portal. I played around with the padding but report bug's position still wasn't shifting, hence used <br>, as it was the shortest way of doing so. Pls review it!

chutten (commented):
The reason padding doesn't work on `form` or `a` is because they are inline. The reason `form` is too high is because it is using `float` to position itself to the right, and the padding space inside the `header` is available for the `float` to use.

I think we're having problems getting this to work because of how we have to bend over backwards to make the `header` escape `.container-fluid`. We should pull `header` out to the root since that's how we want it to behave. Then this should become easier to do.

What do you think?

shreya99oak:
I tried it out, didn't have to set the left and right margin negative, so that works! The only issue is with the top, where adding a padding:top shifts the "report bug" up cause it's an inline element,as you pointed out.

shreya99oak:
@chutten, Since <a> is inline can we set an id to report bug and write it's display as block? I tried it, it works. We just have to set a greater value(let's say 20px) to this id's padding-top than that of .portal-header-bar(say 10px)

chutten:
If we're going to the extent of making it a block, we can use flexbox to align things sensibly. That gives us more control than `float`.

shreya99oak:
@chutten, Fixed it!...https://shreya99oak.github.io/telemetry-dashboard-probe/

index.html
```diff
@@ -9,22 +9,21 @@
...
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css">
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
-
     <link rel="stylesheet" href="new-pipeline/style/dashboards.css">
     <script async src="https://www.googletagmanager.com/gtag/js?id=UA-49796218-2"></script>
 </head>
 <body>
-    <div class="container-fluid">
-        <header>
+        <header class="portal-header-bar">
```
chutten: You'll want to pull `<header>` and all of its contents back to the same indentation level as `.container-fluid` to show that they're now siblings.

shreya99oak: okay

shreya99oak:
@chutten, made the changes

chutten (approved):
Excellent, thank you.

chutten:
Oh dear, this branch has merge conflicts. Could you rebase and repush this so it can be landed?

shreya99oak:
@chutten, resolved the merge conflicts

chutten:
According to github's automerge, it still cannot be rebased due to conflicts.

Did you pull in the latest from mozilla/telemetry-dashboard's repo's gh-pages branch? I sometimes update to my fork's `gh-pages` branch, and it's always behind :S

shreya99oak:
Hey @chutten, Yes, it was not updated! After pulling in the changes my gh-pages and clear-heading branch is up-to-date. However when I tried pushing in the changes, this message appeared..Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes....The solution online was to do a force push..but it can be dangerous at times right? How should I go about it?

chutten:
With properly-set-up repositories, any branch you are permitted to force-push to should be safe to force-push to : ) (but this is an ideal case)

The risk in force pushing is that you are rewriting history. For example: The branch you're pushing _to_ could have commits A,B,C, whereas the branch you're pushing _from_ may have commits A,B,D. 

D could just be C with a better commit message, but git doesn't know that or care. It just knows that you've changed "history" so you have to be _really sure_ you want to proceed. Because after you force-push, C is _gone_.

(( It is usually possible to get it back, but it can be confusing if the remote branch you're pushing to is shared amongst several contributors. ))

In this case, you're only pushing to your own `clear-heading` branch. You don't share it with anyone, and the current contents of that branch aren't as important to you as the new contents you're trying to push. So the `-f` should be fine.

shreya99oak:
@chutten Okay, thanks!! :) Did a force push!

chutten:
Where did you force push to? This branch (`clear-heading`) appears to not have changed?

shreya99oak:
I pushed it to clear-heading @ chutten

chutten:
Hm. Did you rebase on top of mozilla's gh-pages? I still cannot rebase this branch for merge.

shreya99oak:
Yup had done that..what could be the problem?

chutten:
Hrm, interesting, according to the label at the top of https://github.com/shreya99oak/telemetry-dashboard-probe/tree/clear-heading it says "This branch is 63 commits ahead, 1 commit behind mozilla:gh-pages. "

Let's try this:

`git add remote moz https://github.com/mozilla/telemetry-dashboard.git` This will add mozilla's repository as a known remote
`git fetch moz` This will update your local state to know what `moz` knows
`git checkout clear-heading` (this makes sure we're on your `clear-heading` branch, and might not be needed)
`git branch --set-upstream-to=moz/gh-pages` This will tell the clear-heading branch to "track" `moz-gh-pages`
`git pull --rebase` This will update against moz again, and will rebase your work on top of its `gh-pages` branch.

georgf:
@shreya99oak Can we help with anything else here?

georgf:
Hi @shreya99oak,

i'll close this PR temporarily, so it doesn't show up in work trackers.
When you are able to work on it, we can reopen it. Just let me know! :)

PR #450, opened 2018-03-07
Fix #443 : Add shortlink button to probe dictionary

Sylvia23:
To make it easier to pass around links to specific searches in the probe dictionary, a button is added to get shortlinks. The button is added in the navbar with title "Get Permalink".
Two files are affected : `index.html` and `explore.js`

PR #471, opened 2018-03-14
Fix #459 : Stats tab doesn't display if channel is set to any

### LINKED ISSUE #459: Stats tab doesn't display if channel is set to "any"
If the channel is set to "any", the stats tab won't load properly due and will log

```console
TypeError: data[0] is undefined
```

This is because `gChannelInfo.any.versions` is empty, which appears to be because `extractChannelInfos` doesn't put any version information in.

We should handle this somehow... but I'm not sure how. @georgf, do you have an idea about how this could be fixed ideally?

Sylvia23:
If the channel is set to "any", the stats tab won't load properly due and will log
TypeError: data[0] is undefined
Redirecting the channel to proper page fixes the issue.

### DIFF
--- a/probe-dictionary/explore.js
+++ b/probe-dictionary/explore.js
@@ -185,8 +185,8 @@ function showSearchOnlyFilters(show) {
     }
 
     var channel = $("#select_channel").val();
-    if (!["release", "beta", "nightly", "any"].includes(channel)) {
-      $("#select_channel").val("any");
+    if (!["release", "beta", "nightly"].includes(channel)) {
+      $("#select_channel").val("release");
     }
   }
 }


probe-dictionary/explore.js
```diff
@@ -172,7 +172,7 @@ function showSearchOnlyFilters(show) {
     "#version-selection-element",
     "#optout-selection-element",
     "#text-search-element",
-    "#select_channel option[value=any]"
+    "#select_channel option[value=release]"
```
chutten: Won't this hide the "any" choice from the search tab? It's a very helpful option, and should remain for search.

probe-dictionary/explore.js
```diff
@@ -185,8 +185,8 @@ function showSearchOnlyFilters(show) {
     }
 
     var channel = $("#select_channel").val();
-    if (!["release", "beta", "nightly", "any"].includes(channel)) {
-      $("#select_channel").val("any");
+    if (!["release", "beta", "nightly"].includes(channel)) {
+      $("#select_channel").val("release");
```
chutten, line 189: We should only be defaulting to release when switching to the stats (ie, non-search) view. This looks like it will always switch to release, regardless of tab.

probe-dictionary/explore.js
Sylvia23: @chutten , Though it doesn't hide `any` choice from the drop-down. But I have still set the `value` to `any` because it's still working without changing it.

probe-dictionary/explore.js
Sylvia23: @chutten, No, the results are displayed for that channel which is chosen from the drop-down except for `any` for which `release` channel data is shown.
Did I get your question right? Does this sound clear? Please correct me if I am wrong. Thanks!

chutten: It sounds like it might do what I describe, but the only way I can convince myself is if there's a live example I can play with. Can you push this to your github pages so I can play with it? (or is your gh-pages branch already full with something else I've asked for :S )

Sylvia23: @chutten , Yes, it works the same what the issue asks for. Can you please review #450 . After that, I will be able to make it live. Thanks!

Sylvia23:
@chutten , Here is the live link : [sylvia23.github.io/telemetry-dashboard/probe-dictionary](url).
Please check. Thanks a lot !

chutten (approved):

chutten:
This works very well and seems to fix the issue. Thank you!

PR #481, opened 2018-03-19
FIX #448 Make the probe-dictionary search results header sticky

### LINKED ISSUE #448: Make the probe-dictionary search results header sticky
In the [probe-dictionary search results](https://telemetry.mozilla.org/probe-dictionary/), it would be great if the table header is always in view, no matter where you scroll to.

Chris mentioned that [`position: sticky`](https://developer.mozilla.org/en-US/docs/Web/CSS/position) might help here.

gargshruti30:
FIX #448 Make the probe-dictionary search results header sticky.
This PR solves this issue .
I have made the changes in a new branch 'sticky'.
@chutten Please review this PR.


probe-dictionary/explorer.css
```diff
@@ -131,3 +131,8 @@ body {
 .nav-item {
     font-weight: bold;
 }
+th{
```
chutten: The style in this file is to have a full line of whitespace between rulesets, and a space between the selector and `{`

probe-dictionary/explorer.css
```diff
@@ -131,3 +131,8 @@ body {
...
+    position: sticky;
+    top: -1px;
```
chutten: Why `-1`? Would `0` not make more sense, or is it convention?

chutten (changes requested):

probe-dictionary/explorer.css
gargshruti30: @chutten I have corrected it.

probe-dictionary/explorer.css
gargshruti30: @chutten It is because when I put it 0 there was some visible space left.

gargshruti30:
@chutten Can you please now review this PR.

chutten (approved):
Excellent, thank you for your contribution.

PR #483, opened 2018-03-19
Issue #440: Create button-group for measurement & evolution dashboards

ecomerford:
@chutten it looks like the auto-copy is now working in Firefox! Could you take a look at these [changes](https://ecomerford.github.io/telemetry-dashboard/) when you get a chance?

Thanks for your help with this one!

### DIFF
--- a/new-pipeline/src/dashboards.js
+++ b/new-pipeline/src/dashboards.js
@@ -133,12 +133,12 @@ $(document)
         // Workaround for broken selection: http://stackoverflow.com/questions/5797539
         var $this = $(this);
         $this.select()
-          .mouseup(function () {
-           $this.unbind("mouseup");
-           return false;
-           });
+        .mouseup(function () {
+          $this.unbind("mouseup");
+          return false;
+        });
       });
-    $(".permalink-button")
+      $(".permalink-button")
       .click(function () {
         var $this = $(this);
         $.ajax({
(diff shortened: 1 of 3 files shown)


new-pipeline/src/dashboards.js
```diff
@@ -171,7 +164,12 @@ $(document)
         });
         document.execCommand('copy');
       });
-  });
+        // Work around to force the clipboard to hold short URL from permalink button.
```
chutten: This block needs to be a few indentation levels back, I think?

chutten (commented):
Looks good aside from indentation.

ecomerford:
Yikes, sorry! I think I need new glasses. The indents should be good now.

chutten (commented):
I think in the past someone may have messed up the indentation in this area of the code. It doesn't have to be fixed here so long as the code we've added hasn't made things worse.

Ensure the copy event listener's at the right level and we can move on to more important things that historical indentation habits :S

chutten:
Pull that copy handler to the left and I think we'll be done here.

ecomerford:
@chutten Should it be in line with the permalink-button selector?

chutten:
Yes. They are both happening in the same block (it's hard to tell, given how messy things have gotten, but it's true) so they should have the same indentation.

ecomerford:
Okay, I see that now. Should the final closing curly bracket be moved then, as well?

chutten:
Yes. It is closing the block started way back on line 2.

ecomerford:
Does this look correct?

new-pipeline/src/dashboards.js
```diff
@@ -161,17 +154,22 @@ $(document)
...
+            .find("input")
+            .show()
+            .val(shortUrl)
+            .focus();
           },
           async:false
         });
         document.execCommand('copy');
       });
-  });
```
chutten: Does this code run? It looks as though you've dropped a closing sequence here.

ecomerford: I did. I've added it back in with a note because I think that's what was confusing me. I kept the event handler within the ready doc function - I think that's appropriate? Could you take a look and let me know?

Thanks!

chutten (commented):
There's one last thing we can do to minimize how many lines of code we change with this patch. We can return the mouseup handler back to its original position. 

Then we're only changing the lines we have to change.

I'm sorry this keeps ping-ponging back and forth.

chutten (approved):
This is great, thank you for your patience in this.

PR #500, opened 2018-03-22
Fix #458 : Only show URL params that have non-default values

### LINKED ISSUE #458: Only show URL params that have non-default values
We store a lot of values in the URL params. We could greatly shorten the links if we only use params for non-default values.

Sylvia23:
We store a lot of values in the URL params. We could greatly shorten the links if we only use params for non-default values. Knowledge of work-flow of code-base and JavaScript resolved the issue. 
`explore.js` is affected. 

@chutten @georgf , Please review. Thanks! 
Here is the live link : [https://sylvia23.github.io/telemetry-dashboard/probe-dictionary/](url)

### DIFF
--- a/probe-dictionary/explore.js
+++ b/probe-dictionary/explore.js
@@ -593,10 +593,37 @@ function updateSearchParams(pushState = false) {
     params.detailView = gDetailViewId;
   }
 
+  if (params['search']=="") {
+    delete params.search;
+  }
+  if (params['searchtype']=="in_any") {
+    delete params.searchtype;
+  }
+  if (params['optout']==false) {
+    delete params.optout;
+  }
+  if (params['channel']=="any") {
+    delete params.channel;
+  }
+  if (params['constraint']=="is_in") {
+    delete params.constraint;
+  }
+  if (params['version']=="any") {
+    delete params.version;
+  }
+  if (params['view']=="search-results-view") {
+    delete params.view;
+  }
+
+  let queryString = "";
+  if (Object.keys(params).length) {
+    queryString = "?" + $.param(params);
+  }
+  
   if (!pushState) {
-    window.history.replaceState("", "", "?" + $.param(params));
+    window.history.replaceState({}, "", queryString);
   } else {
-    window.history.pushState("", "", "?" + $.param(params));
+    window.history.pushState({}, "", queryString);
   }
 }
 


chutten (commented):
One tiny thing is we get a "?" for an empty set of parameters (like if you load https://sylvia23.github.io/telemetry-dashboard/probe-dictionary/ directly). I think it would be nice if we didn't have that for the "everything is default" case.

Why are you only removing the defaults for non-push-state parameter changes? It means clicking into a detail view expands the url quite a lot.

It's a pity the default values are baked into the HTML where we can't get at them. Otherwise we could set and check the default values against a single, canonical source of truth. Hopefully anyone who changes default values will happen upon this code and update these as well.

Sylvia23:
@chutten , I have made the suggested changes. Now, it is good to review. Thanks!

Live link : [https://sylvia23.github.io/telemetry-dashboard/probe-dictionary/](url)

probe-dictionary/explore.js
```diff
@@ -593,10 +603,42 @@ function updateSearchParams(pushState = false) {
...
+  }
+  if (params['version']=="any") {
+    delete params.version;
+  }
+  if (params['view']=="search-results-view") {
+    delete params.view;
+  }
+  
   if (!pushState) {
-    window.history.replaceState("", "", "?" + $.param(params));
```
chutten: This is a bit of an ornate way of doing things. I think we can simplify this with something like
```js
let queryString = "";
if (Object.keys(params).length) {
  queryString = "?" + $.param(params);
}
```

What do you think?

Sylvia23: @chutten , Yeah, sounds very right ! I will change it now. Thanks!

Sylvia23:
@chutten , How does it look now? Please review. Thanks!

Live link : [https://sylvia23.github.io/telemetry-dashboard/probe-dictionary/](url)

probe-dictionary/explore.js
```diff
@@ -578,6 +578,16 @@ function loadURIData() {
   }
 }
 
+function objectSize (obj) {
```
chutten: We no longer need objectSize, do we?

probe-dictionary/explore.js
```diff
@@ -593,10 +603,37 @@ function updateSearchParams(pushState = false) {
...
+  }
+
+  let queryString = "";
+  if (Object.keys(params).length) {
+    queryString = "?" + $.param(params);
+  }
+  
   if (!pushState) {
-    window.history.replaceState("", "", "?" + $.param(params));
+    window.history.replaceState({}, "", queryString);
```
chutten, line 624: Did you mean to replace the string with an object in these two cases?

probe-dictionary/explore.js
Sylvia23: Yes, Sorry I forgot to remove that.

probe-dictionary/explore.js
Sylvia23: Actually, I looked for the [JavaScript Set/Update method](https://gist.github.com/excalq/2961415#file-gistfile1-js-L1). Both of the syntax work well. I am not very sure of the difference between the two. But this looked more appealing to me so I put that. Please guide which is more right. Thanks!

Sylvia23:
@chutten , Can you please review this and suggest the changes whenever you are free. Thanks!

chutten (approved):

chutten:
Thanks for this, it will really help people who want to share links to the Probe Dictionary.

PR #427, opened 2018-03-01
Fixed issue-406: Use different styles for measure heading if description is available

Shruti9520:
Fixed issue-406- Use different styles for measure heading if description is available
Files that are changed in this PR-
1. dashboards.js
2. dashboards.css

PR #501, opened 2018-03-22
Fix #406 : Use different styles for measure heading if description is available

### LINKED ISSUE #406: Use different styles for measure heading if description is available

Sylvia23:
Since it is a follow up issue of #418 , it can be easily solved by adding one `if-else` statement each in `dist.js` and `evo.js` files and rendering the value of headings with variable name `metric` and `description` from `dist.js` and `evo.js` respectively.
It also resolves the merge conflicts in #427 
@chutten @georgf , Please have a look. Thanks!
I will make it live after #500 if required.

### DIFF
--- a/new-pipeline/src/dist.js
+++ b/new-pipeline/src/dist.js
@@ -876,7 +876,11 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
     var channel = $("#channel-version").val().split("/")[0]
     var desc = getDescription(metric, channel, description);
     var link = getDescriptionLink(metric, channel, description);
-    $('#dist-caption-text').html(desc);
+    if (metric != desc) {
+      $('#dist-caption-text').html(desc);
+    } else {
+      $('#dist-caption-text').text("");
+    }
     $('#dist-caption-link').html(link);
   } else {
     $('#dist-caption-text').text(""); // Clear the histogram caption
--- a/new-pipeline/src/evo.js
+++ b/new-pipeline/src/evo.js
@@ -328,7 +328,11 @@ $(function () {
               var link = getDescriptionLink(metric, channel, description);
               $("#submissions-title").text(metric + " submissions");
               $("#sample-counts-title").text(metric + " sample counts");
-              $("#evo-caption-text").html(description);
+              if (metric != description) {
+                $("#evo-caption-text").html(description);
+              } else {
+                $('#evo-caption-text').text("");
+              }
               $('#evo-caption-link').html(link);
               $("#selected-key")
                 .trigger("change");


chutten (commented):
I will want to see this live, yes.

For now, though, it looks as though this will result in the probe name being displayed twice for probes that have no description. I don't think that's what georgf wanted in the issue.

Sylvia23:
Ohh I see. But @chutten, it looks from #427 also that this was intended here. Please correct me if I am missing something. Thanks!

Sylvia23:
@georgf , Can you please clarify what is intended in this issue? Thanks!

georgf:
I agree with Chris, we should only show the probe name once.

Sylvia23:
@georgf So am I misunderstanding the issue? Can you please brief me what is required in this issue? Thanks!

Sylvia23:
@georgf  I got it. Please review. Thanks!

Sylvia23:
@georgf , [Here](https://sylvia23.github.io/telemetry-dashboard/new-pipeline/dist.html#!cumulative=0&end_date=2018-03-26&keys=__none__!__none__!__none__&max_channel_version=beta%252F60&measure=SIMPLE_MEASURES_START&min_channel_version=beta%252F57&processType=*&product=Firefox!Fennec&sanitize=1&sort_keys=submissions&start_date=2018-03-12&table=0&trim=1&use_submission_date=0) is the live link to check this. Thanks!

new-pipeline/src/dist.js
```diff
@@ -876,8 +876,14 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
     var channel = $("#channel-version").val().split("/")[0]
     var desc = getDescription(metric, channel, description);
     var link = getDescriptionLink(metric, channel, description);
-    $('#dist-caption-text').html(desc);
-    $('#dist-caption-link').html(link);
+    if (metric!=desc) {
```
georgf: A small note on styling: We use spaces around the operators, e.g. `x != y`.

new-pipeline/src/dist.js
```diff
@@ -876,8 +876,14 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
...
+      $('#dist-caption-text').html(desc);
+      $('#dist-caption-link').html(link);
+    }
+    else {
```
georgf: For styling, we put both braces on the same line, e.g. `} else {`.

georgf (commented):
Thanks Sylvia! The description was probably a bit vague here.
This makes sense, but i think we should still show the "more details" link where available. E.g. [here](https://bit.ly/2uQsFR1).

Sylvia23:
@georgf I have made the suggested changes. Can you please see? [Here](https://sylvia23.github.io/telemetry-dashboard/new-pipeline/dist.html#!cumulative=0&end_date=2018-03-29&keys=__none__!__none__!__none__&max_channel_version=beta%252F60&measure=GC_MS&min_channel_version=beta%252F57&processType=*&product=Firefox!Fennec&sanitize=1&sort_keys=submissions&start_date=2018-03-12&table=0&trim=1&use_submission_date=0) is the live link. Thanks!

PS : This [snippet of code](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/new-pipeline/src/dashboards.js#L1068) gives the `More Details` link.

georgf (approved):
This is a simple change and works well, cheers!

I'd still like the description text and detail link to have a smaller (normal) font size, but we can address that in a separate issue.

Sylvia23:
Okay! Thanks @georgf 
So, I am opening a new issue for this now.  Thanks!

PR #504, opened 2018-03-26
Fix #404 : First Click on the aggregates selector on Evolution dashboard sometimes does not select anything

### LINKED ISSUE #404: First Click on the aggregates selector on Evolution dashboard sometimes does not select anything
This might be related to #395

STR:
    visit [this permalink](https://telemetry.mozilla.org/new-pipeline/evo.html#!aggregates=0&cumulative=0&end_date=2018-01-08&keys=!__none__!__none__&max_channel_version=beta%252F58&measure=TELEMETRY_SEND_FAILURE_TYPE&min_channel_version=beta%252F53&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=2017-11-15&trim=1&use_submission_date=0)
    click on "eOK"
    try to tick "abort"
    Observe (your first click is discarded).
    You can now ckeck the boxes as expected.

This issue was commented [in this comment](https://github.com/mozilla/telemetry-dashboard/issues/395#issuecomment-368305759)

Sylvia23:
First Click on the aggregates selector on Evolution dashboard sometimes does not select anything and get discarded. Knowledge of workflow of the function `updateAggregates` along with hit and trial worked here while solving the issue.
One file `evo.js` is affected. 
@o0Ignition0o @chutten , Please review. Thanks!

### DIFF
--- a/new-pipeline/src/evo.js
+++ b/new-pipeline/src/evo.js
@@ -406,7 +406,7 @@ function updateAggregates(kind, buckets) {
   }
 
   // Load aggregates from state on first load
-  newAggregates = newAggregates.map(entry => entry[0]);
+  newAggregates = newAggregates.map(entry => entry[1]);
   let aggregates = gInitialPageState.aggregates.filter(aggregate => newAggregates.includes(aggregate));
   if (!gLoadedAggregatesFromState && aggregates.length > 0) {
     gLoadedAggregatesFromState = true;


Sylvia23:
It seems that I have pushed some `config-file` from my machine as well. I will check this now. Sorry for the inconvenience. Thanks!

o0Ignition0o:
@Sylvia23 great job, congratulations! :)
This should be mergeable once the .gitconfig file gets checked out.
Any idea why and how this file appeared ? :)

Sylvia23:
@o0Ignition0o , Thank you so much.
Yeah ! I have figured that out. Pressing `Ctrl+H` displays the `gitconfig` file. Now, I will remove it and push again. That would probably solve the problem. Thanks!

Sylvia23:
@o0Ignition0o @chutten Now it's good to review. Please have a look. Thanks!

o0Ignition0o:
This looks great indeed! 
Thanks a lot for your contribution, I'll wait for @chutten who will merge it :)

chutten:
Could you explain in the commit message why this is necessary? Near as I can tell `loadStateFromUrlAndCookie` returns the `entry[0]`-style of aggregate names (lowercase with hyphens) not the `entry[1]`-style of aggregate names (Capital Case, without hyphens), so the original code should work.

I mean, it's obvious from how it's currently broken and how reverting a895dd6 fixes it that it _doesn't_ work, regardless of what it "should" do : ), but this change needs to explain itself, I think.

o0Ignition0o:
It seems like the newAggregates array's first values are indexes, whereas the second are litterals.
From your comment I guess this is probably not what was expected ?
Or maybe you mean that if the url is built correctly (IE with the values instead of the keys) this bug would be fixed ? 
screenshot from 2018-03-27 19-23-03

chutten:
The structure of these arrays of arrays is supposed to be `[id, label]` where "id" is some sort of string, and "label" is another string, used for display.

In the case of categorical histograms (like your example, @o0Ignition0o) the "id" is an index (which explains how indices end up in the url instead of aggregate labels). It should probably construct ids that start with "bucket-" in its `map` call in `updateAggregates`

But none of that adequately explains why clicking on things doesn't work...

chutten:
This particular aspect of the bug was addressed in #553. However, #404 still seems to be at large.

PR #508, opened 2018-03-28
Fix #502 : Use the aggregate values insead of the keys in the Evolution dashboard URL

### LINKED ISSUE #502: Use the aggregate values insead of the keys in the Evolution dashboard URL
As of now, when we pick aggregates on the evolution dashboard, the URL keeps a list of selected keys, instead of values. 

As @chutten mentionned [here](https://github.com/mozilla/telemetry-dashboard/issues/395#issuecomment-368514169) This might be one cause of bug #395 and #404.
If not, this would still increase overall readability of the urls :)

Steps to reproduce : 
- Visit [this permalink](https://telemetry.mozilla.org/new-pipeline/evo.html)
- Click on eOK
- Select other aggregates :
screenshot from 2018-03-24 12-08-16

- Have a look at the url in the navigation bar : ` evo.html#!aggregates=0!2!3!4 ` 

The selected aggregate keys are separated by !.
It would be great if the values would be used instead, eg ` evo.html#!aggregates=eOk!eUnreachable!eChannelOpen ` etc.

Sylvia23:
Use the aggregate values insead of the keys in the Evolution dashboard URL. The selected aggregate keys are separated by !. It would be great if the values would be used instead, eg `evo.html#!aggregates=eOk!eUnreachable!eChannelOpen` etc.
Knowledge of work-flow of the code-base in `evo.js` file resolves the issue.

@o0Ignition0o @chutten @georgf . Please review. Thanks!

### DIFF
--- a/new-pipeline/src/evo.js
+++ b/new-pipeline/src/evo.js
@@ -4,6 +4,7 @@ var gFilters = null,
   gPreviousFilterAllSelected = {};
 var gCurrentLinesMap; // mapping from keyed histogram keys to arrays of aggregate lines (non-keyed histograms have lines stored in the key "")
 var gCurrentKind; // the kind of the current measure, or null if this can't be determined
+let newAggregates;
 
 var gDefaultAggregates = [
   ["median", "Median", function (evolution) {
@@ -384,7 +385,6 @@ var gLoadedAggregatesFromState = false;
 
 function updateAggregates(kind, buckets) {
   gCurrentKind = kind;
-  let newAggregates;
   if (kind === "enumerated") {
     newAggregates = getHumanReadableBucketOptions(kind, buckets);
     multiselectSetOptions($("#aggregates"), newAggregates, [newAggregates[0][0]]);
@@ -406,7 +406,8 @@ function updateAggregates(kind, buckets) {
   }
 
   // Load aggregates from state on first load
-  newAggregates = newAggregates.map(entry => entry[0]);
+  newAggregates = newAggregates.map(entry => entry[1]);
+  gInitialPageState.aggregates = $("#aggregates").val().map(x => newAggregates[x]);
   let aggregates = gInitialPageState.aggregates.filter(aggregate => newAggregates.includes(aggregate));
   if (!gLoadedAggregatesFromState && aggregates.length > 0) {
     gLoadedAggregatesFromState = true;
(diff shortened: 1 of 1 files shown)


o0Ignition0o:
Hey thanks a lot for your contribution :) 

Looking at @chutten 's comment [here](https://github.com/mozilla/telemetry-dashboard/pull/504#issuecomment-376613214) I think I might have made a mistake while writing the issue, it seems like the keys might have to change to "bucket-0!bucket-1!" etc.

Let's wait for chutten and see how he reasons about the url and the aggregates :)

Sylvia23:
@o0Ignition0o @chutten , But here the keys are according to the value of `aggregate` name in the dropdown menu at a particular `measure` value. Is there any particular reason to set it like `bucket-0!bucket-1!`? 

Here is the [live link](https://sylvia23.github.io/telemetry-dashboard/new-pipeline/evo.html) to check this.
Please have a look. Thanks!

o0Ignition0o:
I'm not quite sure just yet, but reflecting on what's going on [here](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/new-pipeline/src/dashboards.js#L231) and [in this function](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/new-pipeline/src/dashboards.js#L711) There might be things I'm not aware of / I'm not really understanding.

The MR you submitted does exactly what was expected though, congratulations for that :) 
I think I've made a mistake and wrote the issue and the requirements wrong.

I don't have enough knowledge about this project to figure it out just yet ^^

Sylvia23:
@chutten @georgf Can you please tell what exactly is intended here and state the issue accordingly ? Thanks!

chutten:
@o0Ignition0o is correct in that we don't want the labels to be in the url, we want the ids. In the case of categorical histograms the ids aren't mapped properly to the corresponding aggregates (due to an overbearing `filter` in `loadStateFromUrlAndCookie`) which causes some problems.

Sylvia23:
Oh I see. So should I rename the aggregate names in the URL to "bucket-0!bucket-1!" etc.
@o0Ignition0o @chutten @georgf  Please confirm. Thanks!

chutten:
Please give it a try and see if it works. I figure we're still in the exploration phase of this bug and need to nail down a few more pieces of information before we can make a proper fix.

chutten:
@Sylvia23 Did this work out?

chutten:
Closing due to inactivity. We can reopen this if it becomes active again in future.

PR #513, opened 2018-03-29
Add "sec" and "gfx" data to the environment

### LINKED ISSUE #32: Extend the `environment.json` file
We should add the missing environment data (see [here](https://searchfox.org/mozilla-central/source/toolkit/components/telemetry/TelemetryEnvironment.jsm)) to the `environment.json` file

Dexterp37:
I tried this locally and the list loads correctly. There are, however, a few design decisions to consider:

- The "gfx.monitors" data is an array, each entry has some properties. I identified each property as `environment/system.gfx.monitors.screenWidth` but `environment/system.gfx.monitors[*].screenWidth` might be a better option.
- Similarly, `environment/system.gfx.features` can have sub-objects, all with the same properties. I identified them like e.g. `environment/system.gfx.features.*.textureSharing`.

This is a clone of mozilla/probe-dictionary#39 that fixes mozilla/probe-dictionary#32.

### DIFF
--- a/probe-dictionary/environment.json
+++ b/probe-dictionary/environment.json
@@ -3364,5 +3364,1076 @@
     },
     "name": "system.hdd.system.revision",
     "type": "environment"
+  },
+  "environment/system.gfx.D2DEnabled": {
+    "history": {
+      "release": [
+        {
+          "description": "Whether or not Direct2D is enabled. This is Windows only.",
+          "bug_numbers": [1122052],
+          "details": {
+            "kind": "boolean"
+          },
+          "expiry_version": "never",
+          "optout": true,
+          "revisions": {
+            "firstVersion": "42",
+            "last": "latest"
+          }
+        }
+      ],
+      "beta": [
+        {
+          "description": "Whether or not Direct2D is enabled. This is Windows only.",
+          "bug_numbers": [1122052],
+          "details": {
+            "kind": "boolean"
+          },
+          "expiry_version": "never",
+          "optout": true,
+          "revisions": {
+            "firstVersion": "42",
+            "last": "latest"
+          }
+        }
+      ],
+      "nightly": [
+        {
+          "description": "Whether or not Direct2D is enabled. This is Windows only.",
+          "bug_numbers": [1122052],
+          "details": {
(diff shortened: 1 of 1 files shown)


probe-dictionary/environment.json
```diff
@@ -3364,5 +3415,1229 @@
...
+            "firstVersion": "51",
+            "last": "latest"
+          }
+        }
+      ]
+    },
+    "name": "system.gfx.ContentBackend",
+    "type": "environment"
+  },
+  "environment/system.gfx.adapters.description": {
```
georgf: I agree, for the arrays and dictionaries we should have something that highlights the structure.
How about something like `environment/system.gfx.adapters[<index>].description`?

For now this is only for display purposes.

probe-dictionary/environment.json
```diff
@@ -3364,5 +3415,1229 @@
...
+            "firstVersion": "42",
+            "last": "latest"
+          }
+        }
+      ]
+    },
+    "name": "system.gfx.features.compositor",
+    "type": "environment"
+  },
+  "environment/system.gfx.features.*.status": {
```
georgf: What about using `environment/system.gfx.features[<key>].status`?

georgf (commented):
Thanks for this!
I mostly want to clear up the naming of arrays and dictionaries.

georgf:
A thought for future work:
We could probably link the various contents of arrays or objects together in the front-end ("this is part of environment/...addons - see all properties?").

probe-dictionary/environment.json
Dexterp37: Works for me! I had to remove the `<>` from the name otherwise it wouldn't render correctly (unless we encoded the "<" in the JSON file, which seemed ugly). I've opted for `environment/system.gfx.adapters[index].description`.

Dexterp37:
@georgf this is RFAL!

georgf (approved):
The format looks good, thanks!
I trust you on the research. If you feel this needs a critical set of eyes, maybe pull in someone other than me.

Dexterp37:
This should be fine! Let me go ahead and merge it!

PR #517, opened 2018-03-30
Add the `addons` and `experiments` sections to `environment.json`

Dexterp37:
This is, finally, the last batch of additions to the probe dictionary. This follows the same convention for arrays/objects of the `gfx` section. It loads fine locally.


georgf (approved):
This looks good, cheers!

Issue #474, opened 2018-03-15
Link to use counter dashboard when displaying USE_COUNTER2_* probes


georgf:
*(Based on local user feedback)*

In the measurement dashboard, when showing a probe like [USE_COUNTER2_JS_ASMJS_PAGE](https://mzl.la/2tUm4Ei), we should show a link to the use-counter dashboard.

For all probes whose name starts with `USE_COUNTER2_`, we should link to the right section of the [use counters dashboard](http://georgf.github.io/usecounters/).
For the example here, this would [go here](http://georgf.github.io/usecounters/#kind=page&group=JS&channel=beta&version=59).

We should show the link right under the probe description, e.g. labeled "View in use counter dashboard".

georgf:
This would affect `dist.html`, `evo.html` and probably `dashboards.js` in `new-pipeline/`.

ecomerford:
@georgf Do you mind if I give this a shot?

chutten:
By all means, please do! (georgf is on vacation for the next little while, so I'll be pretending to be him in the interim)

ecomerford:
@chutten I'm having trouble selecting the text of the "Value being measured" button. 

Upon inspecting the element, it appears that the text is within a button called `.multiselect-selected-text`.  The text it prints out _does_ include the name of the selection (see below).
 
![screenshot-1](https://user-images.githubusercontent.com/35710168/37631505-3f37621a-2ba7-11e8-9483-d03024ca72f0.png)

_But_ when I try to use this to create a variable,
![image](https://user-images.githubusercontent.com/35710168/37631532-5c6d54e8-2ba7-11e8-9b55-99b6f98c16a4.png)

The variable itself doesn't contain the same text, thus my code doesn't work. In the first screenshot above, you can see what was logged from `measureSelection` in the console.

I hope this makes some sense.

chutten:
You might be able to get it with `$("#measure").val()`

Does that help?

ecomerford:
@chutten `$("#measure).val()` logs `null`. That's why I tried to use the classes I found when inspecting the element, but that doesn't appear to be working, either. I may try to start over because my initial approach hasn't panned out.

If you or @georgf have any suggestions for where to start for this one, that'd be a big help!

Thank you! :)

chutten:
Odd, the only time I see `$("#measure").val()` being `null` is during load. And during load we don't need descriptions : )

Maybe it's a matter of timing, then?

[...]

PR #520, opened 2018-03-31
Issue #474: Link to use counter dashboard for selections that include "USE_COUNTER2"

ecomerford:
Hi @georgf , would you mind taking a look at this when you get the chance?

You can see it live [here](https://ecomerford.github.io/telemetry-dashboard/new-pipeline/dist.html#!cumulative=0&end_date=2018-03-08&keys=__none__!__none__!__none__&max_channel_version=nightly%252F60&measure=USE_COUNTER2_CONSOLE_TABLE_DOCUMENT&min_channel_version=nightly%252F57&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=2018-01-22&table=0&trim=1&use_submission_date=0).

The only issue I'm having now is that if there is no data for the selection, the code doesn't create a new link, and if you switch to a non-use counter measure that doesn't have enough data to load a histogram, then the link does not get hidden. 

I'm not sure why this is happening- when I console log the `group` variable it seems to be updating each time regardless of the data. But when I console log the `useCounterLink` variable, that does not appear to be updating if there is too little data.

Do you have any thoughts on this?

Thank you!

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -57,6 +57,9 @@
                         <option value="child">process type</option>
                     </select>
                 </h2>
+                <h4 class="use-counter-link">
+                  <a href="#">View in use counter dashboard.</a>
+                </h4>
                 <hr></hr>
                 <figure class="col-md-12">
                     <div class="caption-div">
--- a/new-pipeline/evo.html
+++ b/new-pipeline/evo.html
@@ -4,12 +4,15 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    
     <title>Evolution Dashboard</title>
+    
     <!-- Bootstrap and supporting libraries -->
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css">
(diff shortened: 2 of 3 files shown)


georgf:
Cheers, that's good functional progress!

I think what happens here is that we hit [this branching](https://github.com/mozilla/telemetry-dashboard/blob/aabe22aee2c7096f1dd3bc83d30be92ecea8ca78/new-pipeline/src/dist.js#L872), so to clear/hide the link you'd have to handle it in this function (similar for evo.js).

With the current HTML structure i think the element fits better next to the `dist-caption-link` element.
How about moving it there and using a `getUsecounterLink()` function for the contents?
That would follow the current pattern.

ecomerford:
Sorry for the delay! I've gone ahead and moved the link to be right below the `dist-caption-link` element.

When you say handle it in "this" function, are you referring to the new function I'm creating or the branching functions in evo.js and dist.js?

georgf:
No worries!

> When you say handle it in "this" function, are you referring to the new function I'm creating or the branching functions in evo.js and dist.js?

Ah, i mean the branching functions. I was thinking we can add this similar to how `getDescriptionLink()` works.

georgf:
I'll close this PR temporarily so it doesn't show up in our tracked work.
We can reopen it any time, just let me know! :)

PR #529, opened 2018-04-09
Fix #525 : Show shorter labels for distribution graph

### LINKED ISSUE #525: Show shorter labels for distribution graph
We currently always show the descriptions as the graph label in the distribution views.
For [probes with long descriptions](https://mzl.la/2q9eW1O), that's not adding value (see image) and we already show the description above the graph.

Instead we can just use the histogram/measure name.
This affects dist.js and will require to change what we pass as the `x_label` in the 3 branches [here](https://github.com/mozilla/telemetry-dashboard/blob/991bdbd7d54b67e2a4c76ec9452e6623e4429bb2/new-pipeline/src/dist.js#L1178).
I assume using `.measure` instead of `.description` should probably be sufficient.

screen shot 2018-04-05 at 20 49 52

Sylvia23:
We currently always show the descriptions as the graph label in the distribution views.
For probes with long descriptions, that's not adding value (see image) and we already show the description above the graph. So, we should use shorter labels below the graph for better UI.
One file `dist.js` is affected.
@georgf Please review.
[Here](https://sylvia23.github.io/telemetry-dashboard/new-pipeline/dist.html#!cumulative=0&end_date=2018-04-02&keys=__none__!__none__!__none__&max_channel_version=beta%252F60&measure=SCALARS_BROWSER.ENGAGEMENT.TAB_OPEN_EVENT_COUNT&min_channel_version=beta%252F57&processType=*&product=Firefox!Fennec&sanitize=1&sort_keys=submissions&start_date=2018-03-12&table=0&trim=1&use_submission_date=0) is the live link.


georgf (approved):
This looks good, thanks!

PR #532, opened 2018-04-12
Fix#526:  Probe dictionary stats should only count "release" probes for release channel

Insiyaa:


probe-dictionary/explore.js
```diff
@@ -967,7 +967,9 @@ function getMeasurementCountsPerVersion() {
           // If so, increase the count.
           if (recording) {
             let k = recording.optout ? "optout" : "optin";
-            data[k] += 1;
+            if (k === "optout") {
```
georgf: I think this means we will only count "optout" probes, no matter what channel is selected.
Have you tried selecting the nightly or beta channel and see if they still show both optin and optout probe counts?

The behavior we want is:
- When the release channel is selected, count only "optout" probes.
- When any other channel is selected, count both "optout" and "optin" probes.

georgf (commented):
Thanks Insiyaa!
Do you have a live link where we could test the changes?

Insiyaa:
Oh I missed that! Made the required changes! Here's the [link](https://insiyaa.github.io/telemetry-dashboard/probe-dictionary/). Please review!

probe-dictionary/explore.js
```diff
@@ -967,7 +967,15 @@ function getMeasurementCountsPerVersion() {
...
           if (recording) {
             let k = recording.optout ? "optout" : "optin";
-            data[k] += 1;
+            // For release channel show only "optout" probes
+            if (channel === "release") {
+              if (k === "optout") {
+                data[k] += 1;
+              }  
+            }
+            else {
```
georgf: I think we can bring this down from 2 `if` and 1 `else` to one single `if` statements if we pick the right logical expression.

georgf (commented):
This looks right now, thanks!
One thing i noticed when testing the live link is that this only addresses the "recorded" stats, not the "expired" and "new" ones [like e.g. here](https://insiyaa.github.io/telemetry-dashboard/probe-dictionary/?optout=true&channel=release&constraint=is_expired&view=stats-view).

probe-dictionary/explore.js
Insiyaa: Yes will do it. I just wanted to confirm my approach...

Insiyaa:
The changes are [live](https://insiyaa.github.io/telemetry-dashboard/probe-dictionary/). Please review:)

Sylvia23:
@Insiyaa  One thing that I noticed. There are two `legends` [here](https://insiyaa.github.io/telemetry-dashboard/probe-dictionary/?optout=true&channel=release&view=stats-view) when the channel is set to `release` where `optin` has no significance. I think that shouldn't be there. Does that sound right?
Thanks!

Insiyaa:
@Sylvia23  Yeah I though of it too. But I'm not sure about implementing this. Like I think I've to set its display to None on some condition, but I'm not able to figure that out. It'd be great if you could hint me the direction:)

Sylvia23:
@Insiyaa  I think you can try `someelement.style.display = none;` method of styling inside `JS`  or `legend.remove()` for condition when channel is set to `release` for `optin` legend in the `explore.js` itself. 
I can give you a more clearer view after doing some `queries` on the `variables`. As of now, I think above should work. I will check and tell you ASAP.
Thanks!

Sylvia23:
Hi @Insiyaa, I was able to solve the issue after understanding the codebase of `explore.js` specially the functions `getMeasurementCountsPerVersion()` and `renderProbeStats()`. We can easily solve the issue by just counting the `optin` and `optout` values [here](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/probe-dictionary/explore.js#L990) by declaring them globally so that we can use the values in the function `renderProbeStats()` as well. 
After doing above steps, an `if-else` condition needs to be put just before the `legend` is declared [here](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/probe-dictionary/explore.js#L1084), to generate legends for only `non-zero` values. Put the below code-snippet [here](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/probe-dictionary/explore.js#L1083).

`if(count_optin == 0) columns = ["optout"];
else if (count_optout == 0) columns = ["optin"];`

This way, legend is only generated for `non-zero` values of the `columns(data)`. 

I have also tested it for all the `channels` like `beta`, `release` etc. on my local machine, it works fine for me. This method solves the legend issue generally, not only for channel `release`. I hope this helps.
Thanks!

Insiyaa:
Thanks a lot @Sylvia23 for taking the time out to guide me!:) Your explanation made it very clear... I'll make the required changes soon!

Sylvia23:
@Insiyaa Cheers !

Insiyaa:
Made the changes! Thanks @Sylvia23!
@georgf please review [here](https://insiyaa.github.io/telemetry-dashboard/probe-dictionary/). :smile:

probe-dictionary/explore.js
```diff
@@ -1085,6 +1096,13 @@ function renderProbeStats() {
       .attr("fill", "#000")
       .text("Count of " + constraintText + " probes");
 
+  if(optin_count == 0) {
```
georgf: Instead of using a global variable, we can check this condition locally without introducing global state.
E.g. we can show the `"optout"` column when the array `data` contains any element where `element.optout > 0`.

A small thing: We generally use a whitespace before parantheses and similar elements, i.e. `if (...`.

probe-dictionary/explore.js
```diff
@@ -1085,6 +1096,13 @@ function renderProbeStats() {
...
+    columns = ["optout"];
```
georgf: One more flexible pattern you can use here is:
```
var columns = [];
if (...) {
  columns.push("optout");
}
...
```

probe-dictionary/explore.js
```diff
@@ -939,6 +942,12 @@ function getMeasurementCountsPerVersion() {
     };
   }
 
+function releaseProbesCount(data, k) {
```
georgf: Nice choice to not repeat the logic and put it in a function instead!

I see that this also counts non-release/non-"optout" probes.
Can we name this more descriptively? E.g. `countProbe()`?

Also a small style issue: This is a function that is defined locally, so it should be indented to the surrounding code.

georgf (commented):
Thanks, this looks good!
I only comment on some smaller issues below.

probe-dictionary/explore.js
```diff
@@ -939,6 +939,12 @@ function getMeasurementCountsPerVersion() {
     };
   }
 
+  function countProbe(data, k) {
+    if ((channel === "release" && k === "optout") || (channel !== "release")) {
```
georgf: I think we can simplify this to `(channel !== "release") || (k === "optout")`.

georgf (commented):
Thanks! There are still some small improvements i'd like to see to code clarity and style.

probe-dictionary/explore.js
```diff
@@ -1002,6 +1008,17 @@ function getMeasurementCountsPerVersion() {
 
 function renderProbeStats() {
   var data = getMeasurementCountsPerVersion();
+  var optin_bool = false;
+  var optout_bool = false;
```
georgf: I have some smaller notes here on the style and approach.

We should use more descriptive names here, for example `hasOptinData` & `hasOptoutData`.

Another thing is that the `data.forEach()` loop always goes through all of the array contents.
The only thing that we need to know is if there is *any* optin or optout data in the arrays.

We can do that more clearly and efficiently in one line, e.g.:
```
var hasOptinData = (data.find(item => item.optout > 0) !== undefined);

File: probe-dictionary/explore.js [hunk 8]
```diff
@@ -1085,6 +1102,17 @@ function renderProbeStats() {
       .attr("fill", "#000")
       .text("Count of " + constraintText + " probes");
 
+  columns = [];
```
georgf: This should be `var columns = ...`, otherwise it [becomes a global variable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#Description).

probe-dictionary/explore.js
```diff
@@ -1085,6 +1102,17 @@ function renderProbeStats() {
...
+  if (!optin_bool) {
```
georgf: We can simplify this here to just two independent `if` statements, e.g.:
```
if (hasOptinData) {
  ...push(...);
}
```

georgf (commented):
I think something i missed actually publishing some comments over the last full week, sorry about that!
I'll comment on the remaining smaller issues below now.
Thanks for staying on this, i know it's been taken a few iterations now.

Testing your live link, i also noticed that the opt-out legend is missing for only the "recorded" filter on "beta" and "nightly".

Insiyaa:
Thanks a lot @georgf for guiding me so patiently! Please review and let me know if any other changes are required...

probe-dictionary/explore.js
```diff
@@ -1085,6 +1094,14 @@ function renderProbeStats() {
       .attr("fill", "#000")
       .text("Count of " + constraintText + " probes");
 
+  var columns = [];
```
georgf: A small thing:
I only now noticed that we already have `var columns` defined on line 1019.
Let's just rename this one to `var legendLabels` or so?

georgf (commented):
Thanks for the persistence Insiyaa, this is nearly there!

When testing this live, i get an error in the console through these steps:
- switch to stats
- select channel "release"
- select filter "new"
- now select channel "nightly"
The web console shows the error "TypeError: data is undefined" and the data doesn't update.

Insiyaa:
The case new_in was incomplete. Please review the fix. Changes are live [here](https://insiyaa.github.io/telemetry-dashboard/probe-dictionary/).

georgf (approved):
This looks good to me now, thank you!

PR #535, opened 2018-04-16
Fix #516 : Update TMO frontend to handle 405s from aggregator

### LINKED ISSUE #516: Update TMO frontend to handle 405s from aggregator
In an upcoming aggregator deploy we've changed the API to return a 405 if we are passed filters that are not supported. The short list is:

    ALLOWED_DIMENSIONS = ('application', 'architecture', 'child', 'dates', 'label', 'metric', 'os', 'version')

If we receive a query string parameter, e.g. `?e10sEnabled=true` the API will return a 405 with the unsupported filter removed in the "Allow" header.

The front-end should probably or telemetry.js should probably bubble this up to the user.

Sylvia23:
Handle of `405` response status in the same way as existing handling of `404` and `500` response status in case of bad query string.
One file `v2/telemetry.js` is affected.
@georgf @robhudson @fbertsch  @chutten 
Please review.
Thanks!

### DIFF
--- a/v2/telemetry.js
+++ b/v2/telemetry.js
@@ -477,8 +477,8 @@
   function populateEntriesMap(entriesMap, url, callback) {
     Telemetry.getJSON(url, function (histograms, status) {
       if (histograms === null) {
-        assert(status === 404 || status === 500 , "Could not obtain evolution: status " +
-          status + " (" + url + ")"); // Only allow null evolution if it is 404 or 500 - if there is no evolution for the given filters
+        assert(status === 404 || status === 500 || status === 405 , "Could not obtain evolution: status " +
+          status + " (" + url + ")"); // Only allow null evolution if it is 404, 500 or 405 - if there is no evolution for the given filters
         callback({});
       } else {
         histograms.data.forEach(function (entry) {


robhudson:
I added an example URL in the issue. I'm not sure (and not the person to tell you) if this code change is the right place to catch any areas where this kind of URL could be generated.

I also noted this change doesn't handle the "Allow" header in the 405 and the reloading of the URL with the corrected URL.

Sylvia23:
Thanks @robhudson . Yes I was able to generate 405 errors with that URL, and this code change catches it.

Regarding reloading the url with "Allow" header url, I was unable to get the "Allow" header in this part of the neither in the original definition of `Telemetry.getJSON` using `getResponseHeader("Allow")`. Maybe @chutten  or @georgf  can help in this matter.

chutten:
Maybe getJSON needs to call the callback with three parameters. The third can be the value of the Allow header in the case of a 405 response. The code would need to go somewhere that has access to the xhr object during the onload event, which would put it somewhere... around [here](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/v2/telemetry.js#L361)?

Sylvia23:
Hey @chutten I mentioned in my [comment](https://github.com/mozilla/telemetry-dashboard/pull/535#issuecomment-382106708), that I am unable get the "Allow" header value (url) neither in the part where the aggregator getJSON is [called](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/v2/telemetry.js#L478) nor even with the `xhr` object during the onload event [here](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/v2/telemetry.js#L361). 

With the `xhr` object [here](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/v2/telemetry.js#L361), I tried using ` this.getResponseHeader("Allow")`/` xhr.getResponseHeader("Allow")` both of which results in undefined, and `this.getAllResponseHeaders()` doesn't include the "Allow" header details (it doesn't even contain all the headers) even though the actual response header contains the "Allow" header with the url. I don't understand why this is happening, though I feel this has something to with the way we are making sending the request with `xhr`. Can you check this? @chutten

chutten:
Oh, I'm sorry, I didn't understand you properly when you said it earlier. That's my fault.

I've poked around in a console and I think I have a theory for what's going on. The reason we can only get one header from the response is that it's a cross-origin response. Cross-origin responses only expose a few headers unless the response contains a `Access-Control-Expose-Headers` header.

Further detail is here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers

So, for now we can't do the redirect. The best we can do for now is handle the 405 appropriately (like your patch does), and file a new Issue for redirecting requests when the Allow header is permitted to be accessed.

@robhudson @fbertsch -- unless you wanna fast-track a patch to the service to add `Access-Control-Expose-Headers: Allow` to the 405 responses?

Sylvia23:
Thanks @chutten .
Then maybe you can review this PR for now and I can open a new issue titled "Redirecting requests when the Allow header is permitted to be accessed" and you can mark it as `blocked` or something like that as of now till the backend works well for it. Does this sound good?
Thanks!

chutten (approved):

chutten:
@Sylvia23 Thank you for this. I look forward to the new issue.

Do you want to file the `Access-Control-Expose-Headers` issue against github.com/mozilla/python_mozaggregator as well?

Sylvia23:
Yes @chutten , I would like to file this issue against [github.com/mozilla/python_mozaggregator](url) as well. Thanks!

PR #538, opened 2018-04-20
Fix #537 : Probe dictionary should not show all probes by default

### LINKED ISSUE #537: Probe dictionary should not show all probes by default
It's a bit annoying to go to the probe-dictionary only to wait a few seconds while everything loads. Instead it should be blank, and searches should show results.

Sylvia23:
It's a bit annoying to go to the probe-dictionary only to wait a few seconds while everything loads. Instead it should be blank, and searches should show results.
Making the loading indicator `non-modal` could solve the problem.
Three files `index.html`, `explore.js`, `explorer.css` are affected.
[Here](https://sylvia23.github.io/telemetry-dashboard/probe-dictionary/) is live link to check this.
@chutten @fbertsch @georgf Please have a look and suggest changes and additions.

### DIFF
--- a/probe-dictionary/explore.js
+++ b/probe-dictionary/explore.js
@@ -143,7 +143,6 @@ $(document).ready(function() {
     let date = new Date(gGeneralData.lastUpdate);
     $("#last-updated-date").text(date.toDateString());
 
-    $("#loading-overlay").addClass("hidden");
     mark("done");
   }, e => {
     console.log("caught", e);
--- a/probe-dictionary/explorer.css
+++ b/probe-dictionary/explorer.css
@@ -34,33 +34,6 @@ body {
     z-index: 15;
 }
 
-#loading-overlay {
-    position: absolute;
-    left: 0;
-    top: 0;
-    bottom: 0;
-    right: 0;
-    background: #000;
-    opacity: 0.7;
-    filter: alpha(opacity=70);
-    z-index: 14;
-}
-
-#loading-overlay .progress {
-    width: 76%;
-    /*height: 10%;*/
-    height: 30px;
-    position: absolute;
-    top: 50%;
-    left: 12%;
-    border: 6px solid grey;
-}
-
-#loading-overlay .progress-bar {
-    width: 100%;
-    height: 100%;
-}
-
 #last-updated {
     color: rgba(255, 255, 255, 0.65);
     font-size: 80%;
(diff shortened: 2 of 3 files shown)


Sylvia23:
I have only removed the loader as of now. Please advise on what is intended next. Any advice on where and how to get started with the implementation is also most welcome! Thanks!

Sylvia23:
@chutten @fbertsch @georgf On searching more, I think this [idea](https://css-tricks.com/snippets/jquery/run-javascript-only-after-entire-page-has-loaded/) might help in solving the issue. Please have a look and suggest.
I have tried it on my local machine. We have to just put the code inside `explore.js` file inside the below mentioned function.

`$(window).bind("load", function() {
   // code here
});`

This way the loading time has decreased  because the page gets loaded before the data is loaded. Should I implement it and make it live to check. Please tell.
Thanks!

georgf:
I agree that the responsiveness of the page could be improved.
The large JSON dataset should be loaded asynchronously though, [from `$(document).ready()`](https://github.com/mozilla/telemetry-dashboard/blob/78a0939c76074e01b31edff0e6d35e5883f4a5ae/probe-dictionary/explore.js#L90).

From the users/product side, having an empty page seems a little daunting. The idea with immediately showing the data is that it enables exploration.
Users immediately get to see data and can start refining it.

We probably need to understand first what exactly the performance problem is, by looking at where we spend time doing what. Firefox has some tools built in like the performance or network monitor:
screen shot 2018-04-23 at 18 59 47

georgf:
From the initial look in the performance monitor, it already takes a long time to reach [document ready](https://api.jquery.com/ready/).

Then we're still loading the large JSON datasets, while showing the progress bar.
Until we loaded that data, users can't interact with the page.

Finally, the actual update of the page seems to take ~300ms in my performance run.

georgf:
I missed reading through #537 before. Now that i did i understand the goal of this better.
I propose we settle the design conversation in #537 first before looking at code here.
Sounds good @Sylvia23 ?

Sylvia23:
Yeah! For sure.

georgf:
Sorry this got blocked Sylvia.
I think the best thing for us to focus on #537 and talk about proposals there.

I'll close this PR temporarily so it doesn't show up in our work tracking.
I'm happy to reopen this any time if needed :)

PR #542, opened 2018-04-30
Specify analytics cookie domain

chutten:
gtag.js documentation allows us to customize the cookie name,
domain, and expiry. The defaults are "_ga" "auto" and "two years"
respectively.

"auto" here means that it will set it on .mozilla.org, which isn't what we
want. So let's restrict it to at most .telemetry.mozilla.org.

### DIFF
--- a/analytics/analytics.js
+++ b/analytics/analytics.js
@@ -23,7 +23,9 @@ if (window.trackingEnabled()) {
   window.dataLayer = window.dataLayer || [];
   function gtag(){dataLayer.push(arguments);}
   gtag('js', new Date());
-  gtag('config', 'UA-49796218-2');
+  gtag('config', 'UA-49796218-2', {
+    'cookie_domain': 'telemetry.mozilla.org',
+  });
 
   // Decorate outbound links from the top frame of whitelisted pages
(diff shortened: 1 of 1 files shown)


georgf:
Hm, is the current behavior necessarily a bad thing?
If i start at mozilla.org, already get a cookie set there, then we re-use it on telemetry.mozilla.org...

Is that an outcome we need to avoid?
Do we have current practices elsewhere that we can follow?

chutten:
If mozilla.org sets a cookie on .mozilla.org, this will have no effect on that cookie. This is for cookies added by us on telemetry.mozilla.org which currently attempt to store themselves on the highest domain: .mozilla.org

We (Firefox Telemetry) don't have any analytics on non-telemetry.mozilla.org pages, so we don't care if our cookie is present elsewhere.

From the data practices docs I can find there are no current practices about cookie domains, just about respecting DNT. This does fall under the "no surprises" mandate a little, though, as users may expect cookies set on telemetry.mozilla.org to set -for- telemetry.mozilla.org.

chutten:
Juice not worth the squeeze.

PR #543, opened 2018-05-10
Fix probe recording range display

georgf:
tdsmith found an issue [with the shown recording in the probe-dictionary](https://github.com/mozilla/probe-scraper/issues/40#issue-311418488).
Investigating this it turned out to be a problem with the frontend.

The changes are up for testing [here](https://georgf.github.io/telemetry-dashboard/probe-dictionary/?search=devtools_toolb).

### DIFF
--- a/probe-dictionary/explore.js
+++ b/probe-dictionary/explore.js
@@ -431,24 +431,26 @@ function shortVersion(v) {
   return v.split(".")[0];
 }
 
-function friendlyRecordingRange(firstVersion, expiry) {
-  if (expiry == "never") {
-    return `from ${firstVersion}`;
-  }
-  return `${firstVersion} to ${parseInt(shortVersion(expiry)) - 1}`;
-}
-
-function friendlyRecordingRangeForState(state, channel) {
-  const firstVersion = getVersionRange(channel, state["revisions"]).first;
-  const expiry = state.expiry_version;
-  return friendlyRecordingRange(firstVersion, expiry);
-}
-
 function friendlyRecordingRangeForHistory(history, channel) {
   const last = array => array[array.length - 1];
-  const firstVersion = getVersionRange(channel, history[0]["revisions"]).first;
+
   const expiry = last(history).expiry_version;
-  return friendlyRecordingRange(firstVersion, expiry);
+  const latestVersion = Math.max.apply(null, Object.keys(gChannelInfo[channel].versions));
+  const firstVersion = getVersionRange(channel, last(history).revisions).first;
+  let lastVersion = getVersionRange(channel, history[0].revisions).last;
+
+  if (expiry == "never" && (lastVersion >= latestVersion)) {
+    return `from ${firstVersion}`;
+  }
+
+  if (expiry != "never") {
+    const expiryVersion = parseInt(shortVersion(expiry));
+    if (lastVersion >= expiryVersion) {
+      lastVersion = expiryVersion - 1;
+    }
+  }
+
+  return `${firstVersion} to ${lastVersion}`;
 }
 
 function renderMeasurements(measurements) {
--- a/probe-dictionary/explore.js
+++ b/probe-dictionary/explore.js
@@ -431,10 +431,10 @@ function shortVersion(v) {
   return v.split(".")[0];
 }
 
-function friendlyRecordingRangeForHistory(history, channel) {
+function friendlyRecordingRangeForHistory(history, channel, filterPrerelease) {
   const last = array => array[array.length - 1];
 
-  if (channel == "release") {
+  if ((channel == "release") && filterPrerelease) {
     history = history.filter(h => h.optout);
   }
 
@@ -471,7 +471,7 @@ function renderMeasurements(measurements) {
     ["name", (d, h, c) => d.name],
     ["type", (d, h, c) => d.type],
     ["population", (d, h, c) => h.optout ? "release" : "prerelease"],
-    ["recorded", (d, h, c, history) => friendlyRecordingRangeForHistory(history, c)],
+    ["recorded", (d, h, c, history) => friendlyRecordingRangeForHistory(history, c, false)],
     // TODO: overflow should cut off
     ["description", (d, h, c) => escapeHtml(h.description)],
   ];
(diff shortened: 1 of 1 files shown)


tdsmith:
Thanks! Looking at DEVTOOLS_TOOLBOX_TIME_ACTIVE_SECONDS, I wonder if reporting it as "release from 30" is the right presentation; it became an opt-out metric in release [for 60](https://bugzilla.mozilla.org/show_bug.cgi?id=1446496).

tdsmith:
(The original issue with the _COUNT probe looks good now!)

georgf:
Good catch!

I addressed that, but now [DEVTOOLS_TOOLBOX_TIME_ACTIVE_SECONDS is showing up as recorded from 59 on release](http://localhost:8000/probe-dictionary/?search=DEVTOOLS_toolbo&detailView=histogram%2FDEVTOOLS_TOOLBOX_TIME_ACTIVE_SECONDS).

This seems to actually come from [the source data](https://probeinfo.telemetry.mozilla.org/firefox/all/main/all_probes), i.e. probe-scraper.

georgf:
Does this look correct otherwise?

georgf:
I found another example where recording ranges are off, [telemetry.discarded.child_events](http://telemetry.mozilla.org/probe-dictionary/?search=telemetry.&detailView=scalar%2Ftelemetry.discarded.child_events).
This is also a problem in the source data.

Per [1369041](https://bugzilla.mozilla.org/show_bug.cgi?id=1369041) this probe should be on 56+ and not be on 55 for release & beta.

georgf:
@tdsmith Can you take a look, with that issue in mind?
I'll probably wrap up this frontend issue and take the rest to a separate backend issue.

tdsmith:
Hmm, sorry to nitpick, but the table view in the staging version lists expired prerelease probes as "recorded: never" now.

Compare the "recorded" columns:
https://georgf.github.io/telemetry-dashboard/probe-dictionary/?search=devtools_toolb vs https://telemetry.mozilla.org/probe-dictionary/?search=devtools_toolb

tdsmith:
DOM_TIMERS_* are another interesting set of probes; they're currently displayed as recorded from "46 to 39" on t.m.o

georgf:
Good catches, cheers!
I fixed those and pushed an update.
DOM_TIMERS_* and DEVTOOLS_TOOLBOX_* seem ok.

@tdsmith Do you want to take another look?

tdsmith:
Looking great!

probe-dictionary/explore.js
```diff
@@ -431,24 +431,34 @@ function shortVersion(v) {
   return v.split(".")[0];
 }
 
-function friendlyRecordingRange(firstVersion, expiry) {
-  if (expiry == "never") {
-    return `from ${firstVersion}`;
+function friendlyRecordingRangeForHistory(history, channel, filterPrerelease) {
```
Dexterp37: This reads much better now! I wish this was just a little more documented, as it's still difficult to parse after not looking at this for a while.

probe-dictionary/explore.js
```diff
@@ -431,24 +431,34 @@ function shortVersion(v) {
...
+  const last = array => array[array.length - 1];
+
+  if ((channel == "release") && filterPrerelease) {
```
Dexterp37: comment nit: why is this fixup needed? That's for bridging the optout/optin to the prerelease/release definitions, right?

probe-dictionary/explore.js
```diff
@@ -431,24 +431,34 @@ function shortVersion(v) {
...
   }
-  return `${firstVersion} to ${parseInt(shortVersion(expiry)) - 1}`;
-}
 
-function friendlyRecordingRangeForState(state, channel) {
-  const firstVersion = getVersionRange(channel, state["revisions"]).first;
-  const expiry = state.expiry_version;
-  return friendlyRecordingRange(firstVersion, expiry);
-}
+  if (history.length == 0) {
```
Dexterp37: Why are we returning `"never"` if there's no history? Shouldn't this just be the case for `expiry == "never"`?

probe-dictionary/explore.js
```diff
@@ -461,7 +471,7 @@ function renderMeasurements(measurements) {
     ["name", (d, h, c) => d.name],
     ["type", (d, h, c) => d.type],
     ["population", (d, h, c) => h.optout ? "release" : "prerelease"],
-    ["recorded", (d, h, c, history) => friendlyRecordingRangeForHistory(history, c)],
+    ["recorded", (d, h, c, history) => friendlyRecordingRangeForHistory(history, c, false)],
```
Dexterp37, line 486: nit: should we add `/* filterPrelrease */` after `false`? It would make it easier to understand without having tor scroll the file back to the function definition

probe-dictionary/explore.js
```diff
@@ -843,7 +853,7 @@ function showDetailViewForId(probeId, channel=$("#select_channel").val()) {
     if ((!history[0].optout && (ch == "release")) || (ch == "aurora")) {
       continue;
     }
-    rangeText.push(`${ch} ${friendlyRecordingRangeForHistory(history, ch)}`);
+    rangeText.push(`${ch} ${friendlyRecordingRangeForHistory(history, ch, true)}`);
```
Dexterp37, line 868: nit: same here, but only if it works in the \` :)

Dexterp37 (approved):
This looks good to me! Consider adding the comments where requested below, but don't block on this if you don't have time right now. It's something that would make it easier for other readers in the future, but doesn't need to happen in this bug, as it's a project-wide effort!

probe-dictionary/explore.js
georgf: The prior filter statement can result in an empty history (consider prerelease-only probes on "release" channel).
That should result in a "never" recording range on that channel.

probe-dictionary/explore.js
georgf: That's a good point, i will add some.

probe-dictionary/explore.js
georgf: On the "release" channel we only want to account for the release probes. The most readable approach seems to just filter out the history parts we don't want to account for.

This is only used for the detail view though; in the search results view we want different behavior - hence the `filterPrerelease` flag.

probe-dictionary/explore.js
georgf: That makes these lines pretty long though, i'm torn on that.

Dexterp37: Feel free to drop this and the one below!

PR #544, opened 2018-05-17
Fix to Issue #474: Link to use counter dashboard when necessary

ecomerford:
Hey @georgf, I think I've gotten pretty close here; the branching seems to be working well now and I think this is the placement you were thinking of (?). Correct me if I'm wrong!

I still can't get the link to disappear when the probe switches to a "non-use counter" measure. Upon refresh or if you switch views, the link will go away, but not immediately. Thoughts...?

You can see my changes live [here](https://ecomerford.github.io/telemetry-dashboard/).

Thanks!

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -57,6 +57,9 @@
                         <option value="child">process type</option>
                     </select>
                 </h2>
+                <h4 class="use-counter-link">
+                  <a href="#">View in use counter dashboard.</a>
+                </h4>
                 <hr></hr>
                 <figure class="col-md-12">
                     <div class="caption-div">
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -59,8 +59,9 @@
                 </h2>
                 <hr></hr>
                 <figure class="col-md-12">
+                    <div id="use-counter-link" style="text-align: center">
+                    </div>
                     <div class="caption-div">
-                        <figcaption id="use-counter-link"></figcaption>
                         <figcaption id="dist-caption-text" class="overflow"></figcaption>
                         <figcaption id="dist-caption-link"></figcaption>
                     </div>
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -60,7 +60,7 @@
                 <hr></hr>
                 <figure class="col-md-12">
                   <div>
-                    <figcaption id="use-counter-link"></figcaption>
+                    <figcaption style="text-align: center" id="use-counter-link"></figcaption>
                   </div>
                   <div class="caption-div">
                     <figcaption id="dist-caption-text" class="overflow"></figcaption>
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -58,11 +58,9 @@
                     </select>
                 </h2>
                 <hr></hr>
-                <h4 class="use-counter-link" style="text-align: center">
-                  <a href="#">View in use counter dashboard.</a>
-                </h4>
                 <figure class="col-md-12">
                     <div class="caption-div">
+                        <figcaption id="use-counter-link"></figcaption>
                         <figcaption id="dist-caption-text" class="overflow"></figcaption>
                         <figcaption id="dist-caption-link"></figcaption>
                     </div>
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -59,91 +59,92 @@
                 </h2>
                 <hr></hr>
                 <figure class="col-md-12">
-                    <div id="use-counter-link" style="text-align: center">
-                    </div>
-                    <div class="caption-div">
-                        <figcaption id="dist-caption-text" class="overflow"></figcaption>
-                        <figcaption id="dist-caption-link"></figcaption>
-                    </div>
-                    <div class="col-md-9" id="plots">
-                        <div class="row">
-                            <div class="col-md-6" style="text-align: center">
-                                <h3><select id="selected-key1" class="multiselect" title="Selected key"></select></h3>
-                                <div style="margin-bottom: 20px; z-index: -1">
-                                    <div id="distribution1"></div>
-                                </div>
-                            </div>
-                            <div class="col-md-6" style="text-align: center">
-                                <h3><select id="selected-key2" class="multiselect" title="Selected key"></select></h3>
-                                <div style="margin-bottom: 20px; z-index: -1">
-                                    <div id="distribution2"></div>
-                                </div>
-                            </div>
-                        </div>
-                        <div class="row">
-                            <div class="col-md-6" style="text-align: center">
-                                <h3><select id="selected-key3" class="multiselect" title="Selected key"></select></h3>
-                                <div style="margin-bottom: 20px; z-index: -1">
-                                    <div id="distribution3"></div>
-                                </div>
-                            </div>
-                            <div class="col-md-6" style="text-align: center">
-                                <h3><select id="selected-key4" class="multiselect" title="Selected key"></select></h3>
-                                <div style="margin-bottom: 20px; z-index: -1">
-                                    <div id="distribution4"></div>
-                                </div>
-                            </div>
-                        </div>
-                    </div>
-                    <div class="col-md-3" style="padding-top: 7em">
-                        <div id="summary">
-                            <dl class="dl-horizontal" style="margin-top: 1em">
-                                <dt title="The type of the histogram (e.g., linear, exponential, Boolean)">Histogram Type</dt><dd id="prop-kind"></dd>
-                                <dt title="Total number of Telemetry pings shown">Ping Count</dt><dd id="prop-submissions"></dd>
-                                <dt title="Total number of measurements shown">Sample Count</dt><dd id="prop-count"></dd>
-                                <dt title="Sum of all sample values inserted into the histogram">Sample Sum</dt><dd id="prop-sum"></dd>
-                                <dt title="Number of distinct dates shown">Number of dates</dt><dd id="prop-dates"></dd>
-                                <dt title="Range of dates currently shown">Selected Dates</dt><dd id="prop-date-range"></dd>
-                            </dl>
-                            <div class="scalar-only" style="margin-top: 50px">
-                                <dl class="dl-horizontal">
-                                    <dt>5th Percentile</dt><dd id="prop-p5"></dd>
-                                    <dt>25th Percentile</dt><dd id="prop-p25"></dd>
-                                    <dt>Median</dt><dd id="prop-p50"></dd>
-                                    <dt>75th Percentile</dt><dd id="prop-p75"></dd>
-                                    <dt>95th Percentile</dt><dd id="prop-p95"></dd>
-                                    <br>
-                                    <div style="font-size: 80%; margin: 0 20%">
-                                        <strong>Notice</strong> percentiles are estimated based on values in the histogram. Values are only guaranteed to be accurate to the nearest bucket.
-                                    </div>
-                                </dl>
-                            </div>
-                        </div>
-                        <div id="stats-comparison" style="margin-top: 50px; margin-bottom: 20px;">
-                            <table>
-                              <thead class="measure-headers">
-                              <tr>
-                              <th></th>
-                              <th title="5th percentile duration">5 %</th>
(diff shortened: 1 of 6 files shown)


new-pipeline/src/dashboards.js
```diff
@@ -1081,6 +1081,23 @@ function getDescriptionLink(metric, channel, description) {
...
     }).text(" More details"));
   }
-
   return link;
 }
+
+function getUseCounterLink(metric, channel, description) {
+  var metricUrl = buildDictionaryURL(metric, channel, description);
+  if (metric.startsWith("USE_COUNTER") !== true) {
+    $(".use-counter-link").text(""); // Clear use counter link;
```
georgf: I would suggest to leave dealing with DOM elements to dist.js & evo.js.
So we could do something like:
- default the use counter link to hidden and add it's label to the HTML.
- return `null` from `getUseCounterLink()`, unless its a `USE_COUNTER*`, in which case we return something (the URL to link to or an `<a>` element).
- in dist.js & evo.js, if `getUseCounterLink()` returns `null`, hide/clear the element. Otherwise update & show the link.

new-pipeline/dist.html
```diff
@@ -59,6 +59,8 @@ <h2 style="font-size: 18px; margin-bottom: 0">
                 </h2>
                 <hr></hr>
                 <figure class="col-md-12">
+                    <div id="use-counter-link" style="text-align: center">
```
georgf: Can we follow the style of dist-caption-link below?
It can be on a new line, but follow the font style and add an icon (maybe simply also `fa-info-circle`?).

georgf (commented):
I added some comments below.
I think this is solved most easily by approaching it in a slightly different way.

new-pipeline/src/dashboards.js
ecomerford: Is there an ideal placement for the if/else statements in dist.js and evo.js? Should I keep them within the `displayHistograms` function and the main `telemetry.init` function, respectively?

georgf: Keeping them close to uses of `getDescriptionLink()` should work.
That function covers somewhat similar needs.

ecomerford: Makes sense! :)

ecomerford:
@georgf thanks for your help! It seems to be working properly now.

new-pipeline/dist.html
```diff
@@ -59,89 +59,92 @@ <h2 style="font-size: 18px; margin-bottom: 0">
                 </h2>
                 <hr></hr>
                 <figure class="col-md-12">
-                    <div class="caption-div">
```
georgf: There are a lot of whitespace or indentation changes in this file.
Can you undo those?

new-pipeline/evo.html
```diff
@@ -59,12 +59,14 @@ <h2 style="font-size: 18px; margin-bottom: 0">
...
                 </h2>
                 <hr></hr>
+                <div>
+                  <figcaption style="text-align: center" id="use-counter-link"></figcaption>
+                </div>
                 <div class="caption-div">
-                    <figcaption id="evo-caption-text" class="overflow"></figcaption>
-                    <figcaption id="evo-caption-link"></figcaption>
-                </div> 
-
```
georgf: Nit: unneeded change.

new-pipeline/dist.html
```diff
@@ -59,89 +59,92 @@ <h2 style="font-size: 18px; margin-bottom: 0">
...
-                            </table>
-                        </div>
-                        <div style="text-align: center">
-                            <a class="btn btn-default btn-sm" id="export-csv" title="Export the current histogram to CSV">Export CSV</a>
-                            <a class="btn btn-default btn-sm" id="export-json" title="Export the current histogram to JSON">Export JSON</a>
-                        </div>
-                    </div>
-                </figure>
+                  <div>
+                    <figcaption style="text-align: center" id="use-counter-link"></figcaption>
```
georgf: I think it would be better to place the link after the `caption-div`.
That way we put the description first, then show additional helpful links after it.

The `style` here should go into a CSS file.

This link, when shown, still has a different style from e.g. `dist-caption-link` (smaller font size & blue text color instead of black).
Can we give it the same style?

new-pipeline/src/dashboards.js
```diff
@@ -1081,6 +1081,23 @@ function getDescriptionLink(metric, channel, description) {
       class: "btn btn-outline-primary fa fa-info-circle",
     }).text(" More details"));
   }
-
   return link;
 }
+
+function getUseCounterLink(metric, channel, description) {
+  if (metric.startsWith("USE_COUNTER") !== true) {
```
georgf: A code style nit: `startsWith()` always returns `true` or `false`. We should just use logical tests here, we don't need to check for a value, i.e. `if (!foo) ...`.

We also only support the current use counters, which should start with `USE_COUNTER2_`.

new-pipeline/src/dashboards.js
```diff
@@ -1081,6 +1081,23 @@ function getDescriptionLink(metric, channel, description) {
...
+    return null;
+  }
+  else {
```
georgf: Code style nit: We already `return` above, using an "early return". So we can avoid the indentation here and put the lines below outside of the `else` block, making the function a bit more easy to read.

new-pipeline/src/dashboards.js
```diff
@@ -1081,6 +1081,23 @@ function getDescriptionLink(metric, channel, description) {
...
+    metricSplit = metric.split("_");
```
georgf: Add a `var`, otherwise `metricSplit` is a [global variable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#Description).

new-pipeline/src/dashboards.js
```diff
@@ -1081,6 +1081,23 @@ function getDescriptionLink(metric, channel, description) {
...
+    var group = metricSplit[2];
```
georgf: What happens if the function gets e.g. passed a `metric` string "USE_COUNTER"?

new-pipeline/src/dashboards.js
```diff
@@ -1081,6 +1081,23 @@ function getDescriptionLink(metric, channel, description) {
       class: "btn btn-outline-primary fa fa-info-circle",
     }).text(" More details"));
   }
-
```
georgf, line 1084: Unneeded change?

new-pipeline/src/dist.js
```diff
@@ -869,15 +869,22 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
...
     $('#dist-caption-link').html(link);
+    $('#use-counter-link').html(useCounterLink);
+
   } else {
     $('#dist-caption-text').text(""); // Clear the histogram caption
     $('#dist-caption-link').text(""); // Clear the histogram link
+    $('#use-counter-link').text(""); // Clear use counter link
+  }
+
+  if (!useCounterLink) {
```
georgf: I think we don't need the code here.
This should work with just the changes above.

new-pipeline/src/dist.js
```diff
@@ -869,15 +869,22 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim) {
     var channel = $("#channel-version").val().split("/")[0]
     var desc = getDescription(metric, channel, description);
     var link = getDescriptionLink(metric, channel, description);
-    if (metric != desc) {
```
georgf: Removing this changes the existing behavior.
I think we should just add all the use counter link handling outside of this  `if (metric != desc) ... else ...` , after the `$('#dist-caption-link').html(link);`.

new-pipeline/src/evo.js
```diff
@@ -324,16 +324,24 @@ $(function () {
...
               var description = getDescription(metric, channel, evolutionDescription);
               var link = getDescriptionLink(metric, channel, description);
+              var useCounterLink = getUseCounterLink(metric, channel, description);
               $("#submissions-title").text(metric + " submissions");
               $("#sample-counts-title").text(metric + " sample counts");
               if (metric != description) {
                 $("#evo-caption-text").html(description);
               } else {
                 $('#evo-caption-text').text("");
+                $('#use-counter-link').text("");
```
georgf: This should not be needed as you're already clearing or setting the link text below.

georgf (commented):
This works pretty well now.
I think we can still clean up the code a bit below though.
There are also some whitespace / indentation changes that seem unrelated to the issue here, so they can be dropped.

new-pipeline/src/dashboards.js
ecomerford: Hm. Could that happen? I suppose I can add a safeguard... something like:
```
if (metricSplit.length < 3) {
return null;
}
```

new-pipeline/src/dashboards.js
ecomerford: I'm confused why this didn't go away. Should there be an extra space between the closing bracket on 1083 and the `return link;` line?

georgf: I see a small whitespace on the line below.
It looks like you may have added two space characters on this line?
You can also check this locally by looking at the git diff in the command line or in the Github application.

new-pipeline/src/dashboards.js
```diff
@@ -1081,6 +1081,24 @@ function getDescriptionLink(metric, channel, description) {
...
 }
+
+function getUseCounterLink(metric, channel, description) {
+  var metricUrl = buildDictionaryURL(metric, channel, description);
+  if (!metric.startsWith("USE_COUNTER2")) {
+    return null; // Clear use counter link;
+  }
+  //Show correct use counter link based on group selection.
+  var metricSplit = metric.split("_");
+  var group = metricSplit[2];
```
georgf: `metricSplit` may have fewer than 3 entries. Let's add a check and early return?

new-pipeline/src/dashboards.js
```diff
@@ -1081,6 +1081,24 @@ function getDescriptionLink(metric, channel, description) {
...
+  var useCounterLink = $("<a>", {
+    href: "http://georgf.github.io/usecounters/#kind=page&group=" + group + "&channel=beta&version=60",
```
georgf: Let's just leave out the `version` part, then the usecounters dashboard will always show the latest versions data by default.

new-pipeline/src/evo.js
```diff
@@ -334,6 +335,12 @@ $(function () {
               $('#evo-caption-link').html(link);
               $("#selected-key")
                 .trigger("change");
+              // Clear use counter link when not applicable.
+              if (!useCounterLink) {
```
georgf: Can we move the `var useCounterLink ...` to the line just before this one?

new-pipeline/style/dashboards.css
```diff
@@ -139,7 +139,7 @@
   text-align: center;
 }
 
-#dist-caption-link, #evo-caption-link {
+#dist-caption-link, #evo-caption-link, #use-counter-link {
```
georgf, line 142: Hm, looking at your [live link](https://ecomerford.github.io/telemetry-dashboard/new-pipeline/dist.html#!cumulative=0&end_date=2018-05-24&keys=__none__!__none__!__none__&max_channel_version=nightly%252F62&measure=USE_COUNTER2_PUSHMANAGER_SUBSCRIBE_DOCUMENT&min_channel_version=null&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=2018-05-07&table=0&trim=1&use_submission_date=0), this does not show up with a black text color.

georgf (commented):
Cheers, this looks pretty good!

I just have some style comments below and a small visual issue.

new-pipeline/style/dashboards.css
ecomerford: I added the css key back into the link generator in `dashboard.js`. This is how it's handled for the "More information" link, as well. I don't know exactly why, but that was the only way that I could change the color.

new-pipeline/src/dashboards.js
ecomerford: I think I've fixed this now? I removed the space characters, but your comment wasn't hidden so I'm not sure.

georgf (approved):
This looks good now, cheers!

PR #545, opened 2018-05-18
fix#539: Use standard naming pattern in probe-dictionary code

Insiyaa:


georgf (commented):
This looks good, thanks Insiyaa!
Do you have a live link where we could test this?

Insiyaa:
Yes. It's live [here](https://insiyaa.github.io/telemetry-dashboard/probe-dictionary/).

georgf (approved):
I tested this and it looks good, thanks!

PR #552, opened 2018-06-05
Update mission control URL

wlach:
Mission Control is now in production so we should point people at its new official location.


georgf (approved):
\o/

PR #555, opened 2018-06-15
Refactor document ready handlers

georgf:
*(not important to review during SF allhands)*

We had two document.ready handlers.
This moves the shortlink handlers into the main page initialization
code.
Also clean up that function a little to be less of a monolithic monster.


chutten (approved):
LGTM. 

What's the policy for merging back to its own repo?

Issue #536, opened 2018-04-19
Probe dictionary gives confusing count of probes


jryans:
If you [search for "trigger"](https://telemetry.mozilla.org/probe-dictionary/?search=trigger&searchtype=in_name) in the dictionary, you see confusing output:

![image](https://user-images.githubusercontent.com/279572/39011826-32ad7ed2-43d8-11e8-975f-6645e851d151.png)

It says "Found 3 probes" but only 1 row appears.

It seems this is tied to filters somehow. If I change "channel any" to "channel nightly":

![image](https://user-images.githubusercontent.com/279572/39011878-5763431a-43d8-11e8-8199-283f1dbf7736.png)

...all 3 appear.

Maybe the UI should be more explicit that there were text matches, but they are hidden by channel filters?

chutten:
Huh, "any" should explicitly _include_ nightly. There should be no change in the display between the two screenshots. Both should have those three.

This will likely prove to be an interesting bug.

abhinandan1995:
So i was looking into this issue. @chutten And i can't seem to repro it. Is it still an issue?

chutten:
I'm not sure. It has been a little while since it was filed, so this is entirely possible.

Some possible next steps would be:

1) Ask the person who reported it for help reproducing it
2) Look around merged pull requests here and in mozilla/probe-dictionary to see if there's anything obvious that may have fixed it in the interim
3) Find a new thing to work on while you wait for 1) to happen : )

jryans:
This issue still exists, but you can't reproduce anymore with the original steps. The issue only exists for probes that have only been seen for some but not all channels, which changes over time. For example, when a new probe is first added, it is only recorded by the nightly channel at first, but eventually makes its way to all channels.

As an example that works right now, [search for "DOCGROUPS"](https://telemetry.mozilla.org/probe-dictionary/?search=DOCGROUPS&searchtype=in_name). Channel any says "found 4 probes" but only lists 2.

georgf:
@jryans I have a fix up over at #560, does that look good to you?
[Here is a live instance](https://georgf.github.io/telemetry-dashboard/probe-dictionary/?search=docgroup) of the fixed probe dictionary.

jryans:
@georgf Thanks, that does seem to fix this issue! 😁 

While testing your fixed instance, I noticed the following:

[...]

PR #560, opened 2018-07-04
Fix search results display of probes not yet on release

georgf:
When initially adding the ability to search in "any" channel to the probe dictionary, i used the hack of always only showing the release channel search results.
This breaks when a probe state hasn't propagated yet to release, see #536.

This fixes how we render search results by always showing the “best
availabe” history from “release”, “beta” or “nightly” for each probe.
That avoids hiding probes that haven’t reached the release channel yet
when searching with channel “any”.

See [here](https://georgf.github.io/telemetry-dashboard/probe-dictionary/?search=docgroup) for a live link of the changes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -40,16 +40,20 @@
                         SQL Portal <span class="fa fa-lock"></span>
                         <div class="dashboard-description">View dashboards &amp; run SQL queries.</div>
                     </a>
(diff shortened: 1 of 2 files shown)


Dexterp37 (approved):
This is great, it finally fixes the prerelease channels probes! kudos!

PR #561, opened 2018-07-06
Fix expiry handling in probe dictionary

georgf:
This is live [here](https://georgf.github.io/telemetry-dashboard/probe-dictionary/).

This addresses mozilla/probe-dictionary#41:
- I fixed the recording range that is shown in the search results view, for when a probe will expire.
- To help with understanding expiry, i added human-readable expiry information to the detail view.

### DIFF
--- a/probe-dictionary/explore.js
+++ b/probe-dictionary/explore.js
@@ -465,7 +465,7 @@ function friendlyRecordingRangeForHistory(history, channel, filterPrerelease) {
 
   if (expiry != "never") {
     const expiryVersion = parseInt(shortVersion(expiry));
-    if (lastVersion >= expiryVersion) {
+    if ((lastVersion >= expiryVersion) || (lastVersion >= latestVersion)) {
       lastVersion = expiryVersion - 1;
     }
   }
--- a/probe-dictionary/explore.js
+++ b/probe-dictionary/explore.js
@@ -431,6 +431,10 @@ function shortVersion(v) {
   return v.split(".")[0];
 }
 
+function getLatestVersion(channel) {
+  return Math.max.apply(null, Object.keys(gChannelInfo[channel].versions));
+}
+
 /**
  * Return a human readable description of from when to when a probe is recorded.
  * This can return "never", "from X to Y" or "from X" for non-expiring probes.
@@ -455,7 +459,7 @@ function friendlyRecordingRangeForHistory(history, channel, filterPrerelease) {
   }
 
   const expiry = last(history).expiry_version;
-  const latestVersion = Math.max.apply(null, Object.keys(gChannelInfo[channel].versions));
+  const latestVersion = getLatestVersion(channel);
   const firstVersion = getVersionRange(channel, last(history).revisions).first;
   let lastVersion = getVersionRange(channel, history[0].revisions).last;
 
@@ -471,6 +475,19 @@ function friendlyRecordingRangeForHistory(history, channel, filterPrerelease) {
   }
 
   return `${firstVersion} to ${lastVersion}`;
+}
+
+function friendlyExpiryDescriptionForHistory(history, channel) {
+  let expiry = history[0].expiry_version || "never";
+  if (expiry == "never") {
+    return "never expires";
+  }
+
+  const expiryVersion = parseInt(shortVersion(expiry));
+  let latestVersion = getLatestVersion(channel);
+  let alreadyExpired = (latestVersion >= expiryVersion);
+
+  return `${alreadyExpired ? "stopped" : "stops"} recording in ${expiry}`;
 }
 
 function renderMeasurements(measurements) {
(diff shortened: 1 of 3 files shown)


Dexterp37:
Mh, I think there's something weird going on, or at least some other edge case. See [here](https://bit.ly/2u9225X).

Basically, it's saying:

> Recorded in versions: **beta 43 to 48** / nightly 44 to 49
> Expiry: **beta stops recording in 50** / nightly stops recording in 50

*Expiry* should have said *beta stops recording in 49*, right?

chutten:
It might be English's fault. "stops recording in" is ambiguous of whether it's sometime _during_ that version it stops, or whether it stops _as_ that version begins.

probe-dictionary/explore.js
```diff
@@ -861,13 +861,20 @@ function showDetailViewForId(probeId, channel=$("#select_channel").val()) {
...
+  let expiryText = [];
   for (let [ch, history] of Object.entries(probe.history)) {
     if ((!history[0].optout && (ch == "release")) || (ch == "aurora")) {
       continue;
     }
+
     rangeText.push(`${ch} ${friendlyRecordingRangeForHistory(history, ch, true)}`);
+
+    let expiry = history[0].expiry_version || "never";
+    let expireDetail = (expiry == "never") ? "never expires" : `stops recording in ${expiry}`;
```
chutten: "stops recording in X" is present-tense, which means it'll read strangely for versions in the past like "stops recording in 42". We'll need a tense check based on `latestVersion` or some genius rewrite.

I'm sorry English is so bad at precisely stating negation with continuation.

chutten (changes requested):

georgf:
>> Recorded in versions: **beta 43 to 48** / nightly 44 to 49
>> Expiry: **beta stops recording in 50** / nightly stops recording in 50
>
> *Expiry* should have said *beta stops recording in 49*, right?

Per the probe info data, [this probe](https://georgf.github.io/telemetry-dashboard/probe-dictionary/?detailView=histogram/ABOUT_ACCOUNTS_CONTENT_SERVER_LOADED_TIME_MS) was added in beta 43, removed in beta 48.
The expiry setting though was "expires_in_version: 50", so it would have stopped recording then.

I was thinking it's helpful to expose the expiry information separately and use the "recorded in versions" as the "here is what you actually get". But if our team is confused already by this... :-/

probe-dictionary/explore.js
georgf: Can you suggest alternatives? I'm struggling to find a better, intuitive & precise description.

chutten: If we stick with "stops recording in X" for present/future tense, we can use "stopped recording in X" for past.

Dexterp37:
> I was thinking it's helpful to expose the expiry information separately and use the "recorded in versions" as the "here is what you actually get". But if our team is confused already by this... :-/

Yeah :-\ Tbh, I like the idea of having additional explanation somewhere. Tooltips or some legend somewhere, as this might be a difficult thing to condense in few words, given the confusion is creating.

georgf:
Ok, how about an additional sentence like:
*"This is when the probe will automatically expire and stop recording. Note that the code recording it could be removed before that."*

chutten:
I don't think the confusion resides in whether the probe will be removed before the stated version, though. Introducing that idea probably won't aid understanding of the "does 'expires in' mean 'up to and including' or 'up to but not including'"

georgf:
@Dexterp37 @chutten This is ready for another look.
[This probe](https://georgf.github.io/telemetry-dashboard/probe-dictionary/?search=a11y.&detailView=scalar%2Fa11y.indicator_acted_on) is a good example of both current and future expiry.

Dexterp37 (approved):
I think this is much simpler to understand now, thanks!

chutten:
Unrelated to this patch, the probe says it was recorded in release in 56. I can find no evidence of it being included in anything earlier than 57, though.

Here's where it hits Nightly 58: https://hg.mozilla.org/mozilla-central/rev/905db84ed172
Here's where it hits Beta 57: https://hg.mozilla.org/releases/mozilla-beta/rev/d57f1c698b4e

probe-dictionary/explore.js
```diff
@@ -431,6 +431,10 @@ function shortVersion(v) {
   return v.split(".")[0];
 }
 
+function getLatestVersion(channel) {
+  return Math.max.apply(null, Object.keys(gChannelInfo[channel].versions));
```
chutten: The spread operator (`return Math.max(...gChannelInfo[channel].versions);`) has [excellent cross-browser support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Browser_compatibility) and is more concise.

chutten (approved):
An optional nit.

PR #564, opened 2018-07-11
Tweak error message CSS

georgf:
Tweak the CSS of the error message a bit:
* Add a bit of margin between the error message and the tab elements.
* Also use a bit less padding in the error message as it’s more than
needed.

You can see it live [here](https://georgf.github.io/telemetry-dashboard/new-pipeline/dist.html).


fbertsch (approved):
Nice, I like these changes. Definitely more readable.

PR #566, opened 2018-07-11
Add some tooltips to probe dictionary fields

georgf:
Start adding some tooltips to some probe dictionary fields.
These try to explain some of the potentially less clear fields.

### DIFF
--- a/probe-dictionary/explore.js
+++ b/probe-dictionary/explore.js
@@ -513,13 +513,13 @@ function renderMeasurements(measurements) {
     }],
     ["population", {
       content: (d, h, c) => h.optout ? "release" : "prerelease",
-      tooltip: "Wether this probe collected on Firefox release or only on " +
+      tooltip: "Whether this probe collected on Firefox release or only on " +
                "prerelease channels."
     }],
     ["recorded", {
       content: (d, h, c, history) => friendlyRecordingRangeForHistory(history, c, false),
       tooltip: "What versions this probe is actually recorded in. This depends on " +
-               "when the probe was added, removed and it's expiry.",
+               "when the probe was added, removed and its expiry.",
     }],
     // TODO: overflow should cut off
     ["description", {
--- a/probe-dictionary/explore.js
+++ b/probe-dictionary/explore.js
@@ -502,13 +502,29 @@ function renderMeasurements(measurements) {
   var items = [];
 
   var rawColumns = [
-    ["", (d, h, c) => '<span class="btn btn-outline-secondary btn-sm">+<span>'],
-    ["name", (d, h, c) => d.name],
-    ["type", (d, h, c) => d.type],
-    ["population", (d, h, c) => h.optout ? "release" : "prerelease"],
-    ["recorded", (d, h, c, history) => friendlyRecordingRangeForHistory(history, c, false)],
+    ["", {
+      content: (d, h, c) => '<span class="btn btn-outline-secondary btn-sm">+<span>',
+    }],
+    ["name", {
+      content: (d, h, c) => d.name,
+    }],
+    ["type", {
+      content: (d, h, c) => d.type,
+    }],
+    ["population", {
+      content: (d, h, c) => h.optout ? "release" : "prerelease",
+      tooltip: "Wether this probe collected on Firefox release or only on " +
+               "prerelease channels."
+    }],
+    ["recorded", {
+      content: (d, h, c, history) => friendlyRecordingRangeForHistory(history, c, false),
+      tooltip: "What versions this probe is actually recorded in. This depends on " +
+               "when the probe was added, removed and it's expiry.",
+    }],
     // TODO: overflow should cut off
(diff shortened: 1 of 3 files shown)


georgf:
This is live for testing [here](https://georgf.github.io/telemetry-dashboard/probe-dictionary/).

probe-dictionary/explore.js
```diff
@@ -502,13 +502,29 @@ function renderMeasurements(measurements) {
...
+    }],
+    ["name", {
+      content: (d, h, c) => d.name,
+    }],
+    ["type", {
+      content: (d, h, c) => d.type,
+    }],
+    ["population", {
+      content: (d, h, c) => h.optout ? "release" : "prerelease",
+      tooltip: "Wether this probe collected on Firefox release or only on " +
```
badboy:
```patch
- Wether this probe collected
+ Whether this probe is collected
```

probe-dictionary/explore.js
```diff
@@ -502,13 +502,29 @@ function renderMeasurements(measurements) {
...
+               "prerelease channels."
+    }],
+    ["recorded", {
+      content: (d, h, c, history) => friendlyRecordingRangeForHistory(history, c, false),
+      tooltip: "What versions this probe is actually recorded in. This depends on " +
+               "when the probe was added, removed and it's expiry.",
```
badboy:
```patch
- it's expiry
+ its expire
```

probe-dictionary/index.html
```diff
@@ -156,8 +156,11 @@ <h2 id="detail-probe-name">Probe name</h2>
       <br />
       <table class="table table-sm table-striped table-hover table-bordered border-0">
         <tr><td class="fit pr-2">Type:</td><td id="detail-probe-type" class="grow"></td></tr>
-        <tr><td class="fit pr-2">Population:</td><td id="detail-recording-type" class="grow"></td></tr>
-        <tr id="detail-datasets-row">
+        <tr title="Wether this probe collected on Firefox release or only on prerelease channels.">
```
badboy:
```patch
- Wether this probe collected
+ Whether this probe is collected
```

probe-dictionary/index.html
```diff
@@ -168,7 +171,9 @@ <h2 id="detail-probe-name">Probe name</h2>
         <tr><td class="fit ml-1">Kind:</td><td id="detail-kind" class="grow"></td></tr>
         <tr><td class="fit pr-2">Keyed:</td><td id="detail-keyed" class="grow"></td></tr>
         <tr><td class="fit pr-2">Bug numbers:</td><td id="detail-bug-numbers" class="grow"></td></tr>
-        <tr><td class="fit pr-2">Recorded in versions:</td><td id="detail-recording-range" class="grow"></td></tr>
+        <tr title="What versions this probe is actually recorded in. This depends on when the probe was added, removed and it's expiry.">
```
badboy:
```patch
- it's expiry
+ its expiry
```

Dexterp37:
@georgf , I don't want to sound picky on this, so pardon me :) I was thinking that maybe we should suggest users that there is a "tooltip", otherwise it's not easy to know. What about we add tooltip only on the headers and add a dashed underline where there's a tooltip? As shown [here](https://ux.stackexchange.com/a/80112/19101).

badboy (changes requested):
Some typos for the current version, but I have some other things:

This adds the same text for every probe and is therefore not a more detailed explanation for that specific probe to look at.

Wouldn't it make more sense to change the text per probe to be more specific, e.g. release -> "This probe is only collected on Firefox release channel", pre-release -> "This probe is only collected on Firefox pre-release channels (beta & nightly).

And didn't @chutten propose an alternative saying making it more clear whether the version ranges are including or excluding the last version?

georgf:
>Wouldn't it make more sense to change the text per probe to be more specific, e.g. release -> "This probe is only collected on Firefox release channel", pre-release -> "This probe is only collected on Firefox pre-release channels (beta & nightly)

This sounds great, i'd take that to future improvements though as it's more work.
Having some static tooltips is better than having none.

georgf:
> @georgf , I don't want to sound picky on this, so pardon me :) I was thinking that maybe we should suggest users that there is a "tooltip", otherwise it's not easy to know. What about we add tooltip only on the headers and add a dashed underline where there's a tooltip? As shown here.

It doesn't sound picky & good points.
From user conversations and watching someone try this out, having tooltips around the place of confusion is important.
Underlining seems a bit noisy (visually) to me. I was thinking maybe adding a subtle info circle icon in a follow-up to aid discovery?

Dexterp37:
> Underlining seems a bit noisy (visually) to me. I was thinking maybe adding a subtle info circle icon in a follow-up to aid discovery?

This  sounds like a good compromise to me!

georgf:
@badboy @Dexterp37 ready for another look

Dexterp37 (approved):
Looks good to me!

badboy (approved):

PR #568, opened 2018-07-13
Bug 1472621 Revert Aggregate Data missing banner

jklukas:
The issue with the data is now resolved, so banner no longer needed.


badboy:
:shipit:

badboy (approved):

PR #571, opened 2018-08-08
Configurable evoBucketIndex implementation

ab0092:


dashboard-generator/generator.js
```diff
@@ -265,6 +266,7 @@ window.addEventListener('load', function () {
           sensibleCompare: params['sensibleCompare'][i],
           evoVersions: params['evoVersions'][i],
           filters: params['filters'][i] ? JSON.parse(params['filters'][i]) : '',
+          evoBucketIndex:params['evoBucketIndexes'][i]
```
chutten: needs a space after the ':'

wrapper/README.md
```diff
@@ -22,6 +22,7 @@ Where `params` is:
 * `keyLimit` - a positive integer limiting the number of a keyed measure's measures to be plotted, ordered by number of submissions (default: `4`)
 * `evoVersions` - A number telling us how many versions back to look. If > 0, we will ignore trim, compare, and sensibleCompare and show an evolution instead of a histogram (default: `0`)
 * `percentile` - A number telling us the value below which a given percentage of measurements may be found. This is only valid when evoVersions > 0 (default: `50`)
+* `evoBucketIndex` - The index denoting which bucket index to used to view an evolution for enumerated vendors. (default:0)
```
chutten: Instead of "vendors" I think we should use the word "histograms"

Also, please put a space after the ':'

wrapper/telemetry-wrapper.js
```diff
@@ -353,7 +352,7 @@ window.TelemetryWrapper.go = function (params, element) {
     var percentileLabel = ' - medians'; // i18n?
     var valuesArePercent = false;
     if (kind == 'enumerated' || kind == 'boolean' || kind == 'flag') {
-      const BUCKET_INDEX_FOR_ENUMERATED = 0;
+      const BUCKET_INDEX_FOR_ENUMERATED = params.evoBucketIndex > 0 ? params.evoBucketIndex : 0;
```
chutten: CONSTANT_NAMING_STYLE is something usually reserved for compile-time constants and static configurations. Since we're now configuring this at runtime I think it's worth reworking this to perhaps be called `bucketIndex` or something else written in camelCase so we don't lose that distinction.

dashboard-generator/generator.js
```diff
@@ -82,6 +82,7 @@ window.addEventListener('load', function () {
       sensibleCompare: $('#sensible-compare').checked,
       evoVersions: $('#evo-radio').checked ? $('#evo-versions').value : 0,
       filters: undefined,
+      evoBucketIndex : $('#evo-radio').checked ? 0 : -1
```
chutten: need to remove the space between `evoBucketIndex` and `:`

chutten (changes requested):
Mostly little things, this is looking nicely scoped.

ab0092:
@chutten i have done the requested changes.

wrapper/telemetry-wrapper.js
```diff
@@ -353,14 +352,14 @@ window.TelemetryWrapper.go = function (params, element) {
     var percentileLabel = ' - medians'; // i18n?
     var valuesArePercent = false;
     if (kind == 'enumerated' || kind == 'boolean' || kind == 'flag') {
-      const BUCKET_INDEX_FOR_ENUMERATED = 0;
+      const bucketIndex = params.evoBucketIndex > 0 ? params.evoBucketIndex : 0;
       if (kind == 'boolean') {
         yLabel = desc + ' % FALSE'; // TODO: i18n
```
chutten: Ohhh... here's an interesting case. What if the evoBucketIndex is set to 1 for a boolean histogram? Looks like it'll claim it's "% FALSE" even if we tell it to plot the TRUE bucket.

chutten: Hm, come to think of it... it looks like this should be used for kind == 'boolean' or kind == 'flag'.

wrapper/telemetry-wrapper.js
```diff
@@ -353,14 +352,14 @@ window.TelemetryWrapper.go = function (params, element) {
...
       } else {
-        yLabel = desc + ' - bucket ' + BUCKET_INDEX_FOR_ENUMERATED; // i18n?
+        yLabel = desc + ' - bucket ' + bucketIndex; // i18n?
       }
       valueses = evolutions.map(evo =>
-        evo.map(hist => 100 * hist.values[BUCKET_INDEX_FOR_ENUMERATED] / hist.count
+        evo.map(hist => 100 * hist.values[bucketIndex] / hist.count
```
chutten: This has a possibility of throwing an exception if is `bucketIndex >= hist.values.length`

chutten (approved):
I have found a couple of weaknesses in the code not directly influenced by your changes. We can go ahead and merge things as they are, or if you want we can address them now.

Otherwise it looks as though we can merge this now and file follow-up issues for adding UI on the dashboard generator and other things like that.

What would you prefer?

ab0092:
@chutten the minor issues can be addressed now itself for any major changes we can work on those on a separate item. First for the index out of bound issue we can put a validation check for it, secondly for the bucketindex value for boolean bucket lists what is it that you suggest we go ahead with??

chutten:
Yes, I think validating the index (perhaps when you're checking it for `> 0`on line 355) is a good one to include here. 

I'm also interested to see what you have in mind for fixing the boolean case.

ab0092:
@chutten for the boolean case if only the text of yLabel is to be changed then we can assign to '% TRUE' based on the BucketIndex. For the index validation i think having the check on **line 362** is a better place to validate the index with the hist.values list length.

chutten:
Good point!

ab0092:
@chutten so i will add the index validation and for the boolean instance is my idea good to go with?

chutten:
Yes, please do. Or you could make it so boolean snaps the bucket index to 0, or you can make it so that if it is boolean and non-0 it uses the `yLabel = 'bucket - ' + bucketIndex;` style. There are options to go with, I encourage you to test your preferred one out to make sure it works out well.

ab0092:
@chutten i will go with your suggestion and only set yLabel text to '% False' if the kind is boolean and BucketIndex is 0 else it will show yLabel = 'bucket - ' + bucketIndex;

ab0092:
@chutten i am made the necessary changes please have a look if anything more is to added.

chutten (approved):
Looks good. Do you want to file an Issue for adding UI to dashboard generator while I merge this?

ab0092:
@chutten i think it will be better if you raise the issue instead of me because i am not sure if i will be able to describe the issue as it needs to be done for the other users to understand. I might read the issue and try to understand and may be perhaps add an issue myself later on by taking some reference and assistance from you.

chutten:
Alrighty, I've filed #572. Please take a look and see if it needs any clarification (or if you'd like to work on it!)

ab0092:
@chutten thanks for creating #572. I have gone through it and will work start working on it and provide an update once i have done the necessary changes. Thank you for the assist on resolving this issue.

ab0092:
@chutten just one last assist i require that is when i am creating a PR from my forked repo its still shows my previous approved & merged commits as part of the new Pull Request. Is there any issue from my end while generating the PR ? Thanks..

chutten:
I think maybe it is because you haven't rebased your fork against the latest from `mozilla/gh-pages`. Try adding mozilla's repo as a remote and rebasing your branch against its `gh-pages` branch?

Issue #570, opened 2018-07-18
Hide the "spill" bucket from categorical histograms behind an advanced setting


Dexterp37:
The Measurement Dashboard is showing, for categorical histograms, an additional bucket: the *spill* bucket (see the [example](https://mzl.la/2JxNimJ)).

This bucket is not present on the client side, but it's added [by the aggregator](https://github.com/mozilla/python_moztelemetry/blob/a38e07ebd38a00fadf64b9e068825665dced8849/moztelemetry/histogram.py#L147) as a kind of catch-all / overflow bucket.

However, Firefox simply [discards accumulations](https://searchfox.org/mozilla-central/rev/c296d5b2391c8b37374b118180b64cca66c0aa16/toolkit/components/telemetry/TelemetryHistogram.cpp#2288-2291) for unknown labels.

With that said, I think we should either document the meaning of the *spill* bucket on the dashboard (e.g. tooltip?) so that users are not surprised by it, or remove it. I personally lean toward the second option, unless there's a good reason not to :)

fbertsch:
If the bucket isn't adding any value, it makes sense to not show it. I'm hesitant to make the data unavailable though, because (IIRC) telemetry could accumulate on those buckets if the call is outside of `AccumulateCategorical` and instead uses `Accumulate`. Obviously there is also the case of simply erroneous data.

Maybe we could have an advanced setting to enable showing the spill bucket?

Dexterp37:
> Maybe we could have an advanced setting to enable showing the spill bucket?

Good point on calls outside of the categorical API: we don't forbid that, though it's a weird usage.
I like your proposal, +1 on that!

fbertsch:
Great! Let's plan on that then.

ab0092:
@fbertsch i have looked into the UI. Is it the spill section under success and failure that is to be made visible based on an advanced setting?? Some more details on this issue would be really helpful as i would like to work on this one.

chutten:
Ah yes, this one.

The idea here is that some probes (categorical ones, to be precise) come with one final bucket labelled "spill" that wouldn't really be that useful to people looking at the data on the measurement dashboard. The proposal is that we hide the "spill" bucket on the distribution and evolution views and provide an "Advanced Setting" to allow it to be re-shown just in case someone who knows what they're doing wants to take a look.

Does that help, @ab0092 ?

ab0092:
[...]

PR #574, opened 2018-08-11
 Implemented aggregate values instead of the keys in the Evolution dashboard URL

ab0092:


chutten:
Ah-ha, I knew I was missing something.

Sorry for the delay, I was out of the office last week. I'll take a look at this today.

ab0092:
No issues. Please have a look and let me know if any tweaks are required.. Thanks.

new-pipeline/src/evo.js
```diff
@@ -412,8 +412,14 @@ function updateAggregates(kind, buckets) {
   }
 
   // Load aggregates from state on first load
-  newAggregates = newAggregates.map(entry => entry[0]);
-  let aggregates = gInitialPageState.aggregates.filter(aggregate => newAggregates.includes(aggregate));
+  let aggregates = gInitialPageState.aggregates
```
chutten: Could you provide an explanation for what transformation you're performing here? It seems as though you're taking the initial aggregates and... I'm actually a little lost already.

new-pipeline/src/evo.js
```diff
@@ -412,8 +412,14 @@ function updateAggregates(kind, buckets) {
...
+                    .filter((aggregate)=>{
+                      return newAggregates.filter((newaggregate)=> newaggregate[1] == aggregate).length > 0
```
chutten: This code filters based on aggregates who are present as the second index in one of the newAggregates? Perhaps this will be more clear?

`.filter(aggregate => newAggregates.some(newAggregate => newAggregate[1] === aggregate))`

chutten (commented):
Is it possible to write this code in a backward-compatible manner to ensure that old links people have lying around will still work?

The style in the file is a little eclectic, but mostly we have spaces before and after `=>` and after `,`

Looks like a good first draft!

new-pipeline/src/evo.js
ab0092: The transformation happening here is actually from the values to keys because as we are now storing the values of aggregates in the url instead of the keys so while restoring data back into to UI i have remapped the values to the keys for the multi-select control to work as required.

ab0092: And you are right at the 0th index the keys & at the 1st index the values are stored respectively.

ab0092:
@chutten i have implemented the backward compatibility option and also formatted the code accordingly. Please have a look if anything more is to be added.

new-pipeline/src/dashboards.js
```diff
@@ -227,8 +227,8 @@ function loadStateFromUrlAndCookie() {
     var aggregates = pageState.aggregates.split("!")
       .filter(function (v) {
         return ["5th-percentile", "25th-percentile", "median",
-            "75th-percentile", "95th-percentile", "mean", "submissions"].indexOf(v) >= 0 ||
-          v.startsWith("bucket-");
+            "75th-percentile", "95th-percentile", "mean", "submissions"].indexOf(v) >= 0
+            ||(v.trim() != "");
```
chutten: Need a space after `||`, or you can leave it at the end of the previous line.

new-pipeline/src/evo.js
chutten: I think I get it now. We want a list of newAggregates' keys where either the key or the value of the newAggregate is present in gInitialPageState.aggregates. Is that right? I think we can make this a bit more obvious if we write it something like:
```js
let aggregates = [];
for (let [key, value] of newAggregates) {
  if (gInitialPageState.aggregates.includes(key) || gInitialPageState.aggregates.includes(value)) {
    aggregates.push(key);
  }
}
```

chutten (commented):
Do I have the logical transformation correct here? It's tricky to unwrap the functional transformations needed, so I'm not sure I have it right.

new-pipeline/src/dashboards.js
ab0092: The changes made here is actually done to make the old url's to also work so i have added the change of checking for only non-empty values and later on check the validity of the values in the savestateToCookieandUrl methods

ab0092:
@chutten i have added the required changes. And thanks for the above code snippet idea. It really makes it more understandable and clean.  :-)

new-pipeline/src/evo.js
```diff
@@ -921,10 +926,14 @@ function saveStateToUrlAndCookie() {
     cumulative = gInitialPageState.cumulative,
     trim = gInitialPageState.trim,
     sortKeys = gInitialPageState.sort_keys,
-    selectedKeys = gInitialPageState.keys;
+    selectedKeys = gInitialPageState.keys,
+    aggregateOptions = $("#aggregates option"),
+    // Get Aggregate Values instead of keys to save 
+    // in the state url to increase readability
+    aggregateValues = $("#aggregates").val()
```
chutten, line 933: From the previous code it seems possible that `$("#aggregates").val()` could sometimes fail to return the selected aggregates (thus the need for the `|| []`). Don't we need to check for that case here, too?

chutten (commented):
Looks mostly good, though I think I may have noticed a corner case.

How did you test this? Is there a live preview somewhere I can play with?

new-pipeline/src/evo.js
ab0092: You can check the application here https://ab0092.github.io/telemetry-dashboard/

And for the validation as per my understanding we do not require the extra check as all the items in the multi-select control cannot be unchecked and the first option is always kept checked so may be we do not require it here.

chutten: Interestingly, you are correct. When you remove the final option's checkmark it resets to the initial or default selection (which can be the first aggregate, or some combination of a few of them) which keeps us clear of errors here.

(( this reverting to default selection is probably a bug, but we can file and fix that later. ))

chutten:
Looks good, works well, and I confirm that old-type "bucket-0!bucket-4"-type urls [like this one](https://ab0092.github.io/telemetry-dashboard/new-pipeline/evo.html#!aggregates=bucket-0!bucket-4&cumulative=0&end_date=2018-08-19&keys=!__none__!__none__&max_channel_version=nightly%252F63&measure=TELEMETRY_SEND_FAILURE_TYPE&min_channel_version=nightly%252F60&processType=*&product=Firefox&sanitize=1&sort_keys=submissions&start_date=2018-06-25&trim=1&use_submission_date=0) are compatible with the new logic.

Good work!

chutten (approved):

chutten:
And that should be that. Have an eye on what you'd like to work on next?

ab0092:
@chutten i have not yet been able to look at the open issues in details so could you suggest one which is having an higher priority ?? Or else i will have a look at the open tickets and decide.

chutten:
Take your pick, there are none that are immediately pressing at this moment.

ab0092:
#570 is an issue i had inquired about but did not receive any update so if you could provide me some details then we are good to go. We can discuss the rest on that thread itself.

Issue #575, opened 2018-08-21
Dev edition charts show up as "aurora" in chart labels


robhudson:
Visit https://mzl.la/2BaArYT and hover the chart lines and look at the labels. Since the front-end is already changing "Aurora" to "Dev Edition" elsewhere I propose the front-end does the string swap here as well.

shiri1248:
I would like to work on this.

shiri1248:
can anyone  guide me on this issue please

chutten:
Yes. The code controlling what plots look like can be found in `new-pipeline/src/dist.js` and `new-pipeline/src/evo.js`. Any time it uses the word "aurora" in a place where it might be shown to a user it should instead read "dev edition"

shiri1248:
Thanks @chutten , so I should replace the  word  "aurora" to  "dev edition" am I right?

chutten:
Correct.

shiri1248:
@chutten  submitted a PR #576  please review and  if you find any conflicts please let me  know

hardik25:
@chutten 

How to reproduce this issue?

hardik25:
Hi @chutten 

I have created a pull request with fix to this issue. Please review.

driver-by:
Hi guys,
I've  submitted a PR #629 to fix the issue. Waiting for review.
Thanks!

PR #576, opened 2018-08-24
fixes chart lables

shiri1248:
#575  fixes

### DIFF
--- a/new-pipeline/src/evo.js
+++ b/new-pipeline/src/evo.js
@@ -199,7 +199,7 @@ $(function () {
               var latestChannelVersion = Infinity;
               if (channel === "nightly") {
                 latestChannelVersion = latestNightlyVersion;
-              } else if (channel === "aurora") {
+              } else if (channel === "dev edition") {
                 latestChannelVersion = latestNightlyVersion - 1;
               } else if (channel === "beta") {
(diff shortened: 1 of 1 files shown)


chutten (changes requested):
This code may or may not make a difference as it controls what versions are useful to put in the channel+version dropdowns.

This suggests that the "aurora" labels on the graphs must be coming in from the data. This means intercepting the data just before `MG.data_graphic` is called and mapping "aurora" to "dev edition"

Do you understand what I mean?

shiri1248:
@chutten  thanks for reviewing. I'll take care of it :)

chutten:
@shiri1248 Do you need any help?

shiri1248:
yes @chutten   it would be helpful :)

chutten:
I think the problem lies in the data returned from the aggregator itself by the `Telemetry.getEvolution` call in `evo.js`. It probably has the channel set to "aurora". Luckily, we convert it to `Line` objects before we graph them, so maybe in the `Line` constructor we can take `channelVersion` and replace `aurora` with `dev edition` in it?

Do you think you'd like to give that a try, @shiri1248 ?

shiri1248:
yes @chutten  I will give a try.

chutten:
This PR appears to be abandoned. Closing.

PR #579, opened 2018-09-20
First Auth0 Prototype for Measurement Dashboards

chutten:
By all means give it a try. Head to https://chutten.github.io/telemetry-dashboard/ and get into the Measurement Dashboard. There's a little login button at the top-right. It should redirect you to my auth0 for auth 

user: user2@domain.com
pass: U$erdomain

If you successfully authenticate the login button should disappear and your access token will be at `window.Telemetry.AuthorizationToken`

Limitations: 
* Each navigation needs new auth
  * This includes refreshes and new tabs in addition to navigating between evo and dist
  * Auth0 is smart enough to auth you almost silently, so subsequent logins are just a matter of clicking the login button. No credentials. This may be low-friction enough for MVP, or may not.
* No UI to let users know if they've auth'd or not.
  * The entire design is in need of review, so this isn't much surprise
* Doesn't transmit auth to aggregates.tmo
  * aggregates.tmo actually has some CORS on it saying we can't send any but the most standard of headers. `Authorization` isn't a "most standard" so I had to comment it out.
* No handling for token expiry
  * Tokens are configured to be valid for 2 hours, but I've not implemented any way of renewing it beforehand or handling a case where auth goes wrong.
  * User sessions are short enough and Auth0 authentication is seamless enough that this might be acceptable for MVP.

### DIFF
--- a/new-pipeline/src/auth.js
+++ b/new-pipeline/src/auth.js
@@ -53,9 +53,9 @@ function constructLoginUrl() {
 
 function constructLogoutUrl() {
   let currentUrl = new URL(window.location.href);
-  // Discard the hash in case there's an access_token in it. The dashboard
-  // will restore state from Cookie.
-  currentUrl.hash = "";
+  // Reset the hash to the default measurement on logout since we don't want
+  // lingering release-channel state to restart the login flow.
+  currentUrl.hash = "measure=GC_MS";
 
   let logoutUrl = new URL(AUTH0_ORIGIN + "v2/logout");
   logoutUrl.searchParams.append("returnTo", currentUrl.toString());
--- a/new-pipeline/src/auth.js
+++ b/new-pipeline/src/auth.js
@@ -20,9 +20,9 @@ function randomString(length) {
   return result.join('');
 }
 
-const AUTH0_ORIGIN = "https://chutten.auth0.com/";
-const AGGREGATES_APP_CLIENT_ID = "v427CBip66hS4q2mtJeZJWiYK1aEvQTK";
-const AGGREGATES_API_AUDIENCE_ID = "aggregates.telemetry.mozilla.org";
+const AUTH0_ORIGIN = "https://auth.mozilla.com/";
+const AGGREGATES_APP_CLIENT_ID = "kj04FIzGyg9CjSGphdTelwezVHHV1HVc";
+const AGGREGATES_API_AUDIENCE_ID = "https://aggregates.telemetry.mozilla.org/";
 const AGGREGATES_API_SCOPE = "read:aggregates";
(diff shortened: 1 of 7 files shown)


chutten:
@wlach @mreid-moz  @fbertsch You may find this interesting

new-pipeline/src/auth.js
```diff
@@ -0,0 +1,102 @@
...
+  let loginUrl = new URL(AUTH0_BASE_URL);
+  for (let [key, value] of Object.entries(params)) {
+    loginUrl.searchParams.append(key, value);
+  }
+
+  console.info("Constructed login url: " + loginUrl);
+  return loginUrl.toString();
+}
+
+function createLoginButton() {
```
fbertsch: This should probably be moved to html/css, no?

new-pipeline/src/auth.js
```diff
@@ -0,0 +1,118 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function () {
+
+// The state we used when asking for the token. A 16-character random string.
+// We could use the actual app state if we wanted to.
+let gRequestState;
+
+/* Copied with implicit license from Auth0 docs: randomString */
+function randomString(length) {
```
fbertsch, line 12: There's really no available library function?

new-pipeline/src/auth.js
```diff
@@ -0,0 +1,118 @@
...
+  // lingering release-channel state to restart the login flow.
+  currentUrl.hash = "measure=GC_MS";
+
+  let logoutUrl = new URL(AUTH0_ORIGIN + "v2/logout");
+  logoutUrl.searchParams.append("returnTo", currentUrl.toString());
+
+  return logoutUrl.toString();
+}
+
+function displayAuthButton(buttonText, buttonUrl) {
```
fbertsch, line 66: nit: shouldn't this be defined in the HTML?

fbertsch (approved):
Just a few nits. Let's test this

new-pipeline/src/auth.js
chutten: It can be. Creating it in JS is an artefact from when it was a disappearing login button.

new-pipeline/src/auth.js
chutten: Even if there were, we already have a problem of too many JS files on load.

new-pipeline/src/auth.js
chutten: Actually, upon reflection I'd like to leave it. Messing with individual text nodes to change the button text and href... we can move it to a follow-up

PR #587, opened 2018-11-02
[auth] Use the appropriate moz auth0 domain

chutten:


fbertsch (approved):
plz plz plz plz plz plz work

plzwork

PR #608, opened 2019-01-24
Add a sort-by-value checkbox.

bwinton:
As requested by me, and described by @chutten. 🙂

### DIFF
--- a/new-pipeline/src/dist.js
+++ b/new-pipeline/src/dist.js
@@ -93,8 +93,11 @@ $(function () {
     $("input[name=sanitize-toggle]")
       .prop("checked", gInitialPageState.sanitize !== 0)
       .trigger("change");
-      $("input[name=include-spill-toggle]")
+    $("input[name=include-spill-toggle]")
       .prop("checked", gInitialPageState.include_spill > 0)
+      .trigger("change");
+    $("input[name=sort-by-value]")
+      .prop("checked", gInitialPageState.sort_by_value > 0)
       .trigger("change");
 
     updateOptions(function () {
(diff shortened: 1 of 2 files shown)


new-pipeline/src/dist.js
```diff
@@ -914,6 +918,28 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim, in
     if (histogramsList.length === 1 && histogramsList[0].histograms.length ===
       1) { // Only show one set of axes
       var histogram = histogramsList[0].histograms[0];
+      if (sortByValue) {
+        let sorting = [];
+        for (let index = 0; index < histogram.buckets.length; index++) {
+          sorting.push([histogram.values[index], histogram.buckets[index]]);
+        }
+        sorting.sort((a,b) => {
```
chutten:
```suggestion
        sorting.sort((a, b) => {
```

new-pipeline/src/dist.js
```diff
@@ -914,6 +918,28 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim, in
...
+          rv = b[0] - a[0];
```
chutten: `rv` needs a `let` or `var` or something

new-pipeline/src/dist.js
```diff
@@ -914,6 +918,28 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim, in
...
+          if (rv != 0) {
+            return rv;
+          }
+          if (a[1] < b[1]) {
+            return -1;
+          }
+          if (a[1] > b[1]) {
+            return 1;
+          }
+          return 0;
```
chutten: If you think it's not too difficult to read/maintain we could condense to:

```js
sorting.sort((a, b) => b[0] - a[0] || a[1].localeCompare(b[1]));
```

chutten (approved):
Some nits, but it looks good!

chutten:
Now that I've given a few more moments' thought to it, maybe we should limit this to histograms that are of categorical kind. A linear or exponential histogram should probably _never_ be sorted by value.

new-pipeline/src/dist.js
bwinton: Fixed (kinda).

new-pipeline/src/dist.js
bwinton: Fixed (kinda).

new-pipeline/src/dist.js
bwinton: Changed to this, which removes the previous two nits. 🙂

bwinton:
Something like this latest one?

chutten (approved):

chutten:
Alrighty, it should be up within the next half-hour or so. May need to force-refresh to get through the layers of caching.

Fingers crossed that not everything explodes!

PR #610, opened 2019-02-08
 Fix probe-dictionary event details

georgf:
The event detail view was putting the objects, methods & extra keys into the wrong elements.


chutten (approved):
Why yes, it does seem likely that methods, objects, and extra_keys should go in the respective locations.

PR #615, opened 2019-03-11
Fix for issue #575: Dev edition charts show up as aurora in chart labels

hardik25:
Changing channelVersion to "Dev Edition" from "Aurora" in Line constructor.


chutten:
Looks like this pull request contains changes for two different issues. Could you split this up so we have one PR per issue?

PR #617, opened 2019-03-18
Fix for issue #575: Dev edition charts show up as aurora in chart labels

hardik25:
Submitting patch for fixing issue #575


chutten:
Is this what's currently up at hardik25.github.io? When I navigate to https://bit.ly/2UFQ67H it still says "aurora" on the chart.

PR #620, opened 2019-04-01
Downscope 'copy' event listener

chutten:
No need to squelch all copy events. We only want to manipualte the
ones we added to work around some shortlink problems.

Fixes bug 1519841


chutten:
Fix is live on my fork here: http://chutten.github.io/telemetry-dashboard/

Tested by copying selected body text both before and after using the shortlink. Also by checking that the auto-copy of the shortlink still works.

fbertsch (approved):
Cool, fixed it for me!

PR #578, opened 2018-09-03
Implemented option to Hide the "spill" bucket from categorical histograms behind an advanced setting

ab0092:

PR #624, opened 2019-04-22
Fix index out of bounds when checking spill.

### LINKED ISSUE #606: When selecting a metric that has no data the Dashboard shows the previous data (histogramList[0] is undefined)
In the event that a metric has no data to show (like Nightly 66's DEVTOOLS_FLEXBOX_HIGHLIGHTER_TIME_ACTIVE_SECONDS when "sanitize data" is true), then we error to the log

```console
TypeError: histogramsList[0] is undefined[Learn More] dist.js:813:11
	displayHistograms https://telemetry.mozilla.org/new-pipeline/src/dist.js:813
	<anonymous> https://telemetry.mozilla.org/new-pipeline/src/dist.js:380
dispatch7jQuery	gFilterChangeTimeout https://telemetry.mozilla.org/new-pipeline/src/dist.js:347
	calculateHistograms https://telemetry.mozilla.org/new-pipeline/src/dist.js:570
	updateDateRange https://telemetry.mozilla.org/new-pipeline/src/dist.js:648
	calculateHistograms https://telemetry.mozilla.org/new-pipeline/src/dist.js:568
	Telemetry_getEvolution https://telemetry.mozilla.org/v2/telemetry.js:569
	populateEntriesMap https://telemetry.mozilla.org/v2/telemetry.js:498
	onload https://telemetry.mozilla.org/v2/telemetry.js:380
```

And the view remains on the previous data, whatever it was. We -do- have the ability to show "no data for these options" views via metricsgraphics, so we should probably do that instead of silently throwing a fit.

### LINKED ISSUE #623: Probes that are desktop-only don't provide useful error for mobile on telemetry.m.o
Go to e.g. https://mzl.la/2ZqFJHV and then in the selector sentence at the top, try changing "Firefox Desktop" to "Firefox Mobile".

Expected: the histogram goes away, or gives you some indication that this probe is not recorded on mobile.

Actual: histogram remains pretty much where it is, showing the desktop numbers.

The problem is more obvious if you go to the mobile stats for a different probe, e.g. https://mzl.la/2ZotGL6 and then switch the probe name to FX_TAB_SWITCH_COMPOSITE_E10S_MS - it will keep showing the old probe data.

chutten:
Fixes #606
Fixes #623

Regression from #578

Live demo: https://chutten.github.io/telemetry-dashboard/

### DIFF
--- a/new-pipeline/src/dist.js
+++ b/new-pipeline/src/dist.js
@@ -813,7 +813,7 @@ function displayHistograms(histogramsList, dates, useTable, cumulative, trim, in
   cumulative = cumulative === undefined ? false : cumulative;
   trim = trim === undefined ? true : trim;
   // Hide the 'spill' bucket if present
-  if (!includeSpill) {
+  if (!includeSpill && histogramsList.length) {
     const spillIndex = histogramsList[0].histograms[0].buckets.indexOf("spill");
     histogramsList[0].histograms[0].buckets = histogramsList[0].histograms[0].buckets
                                             .filter((value, index) => index !== spillIndex);


fbertsch (approved):
Two issues, on PR? Nice. 🐦 🐦

PR #627, opened 2019-07-09
bug 1562846 - Don't force all boolean aggregates all the time

chutten:


fbertsch (approved):
This took me a bit longer to understand than it should have :) r+

Issue #625, opened 2019-05-10
Colour graphs the same way on Evolution View


mreid-moz:
From [Bug 1544102](https://bugzilla.mozilla.org/show_bug.cgi?id=1544102): 

> p.s., I find the fact that telemetry colours the three graphs on the evolution view differently to be VERY annoying.

PR #631, opened 2019-11-04
Color aggregate lines the same across versions

jonco3:
At present it can be confusing looking at evolution graphs because the individual aggregates get assigned different line colors for different versions, e.g.:

Screenshot_2019-11-04 Evolution Dashboard

I'm not sure whether that's exactly what this issue is talking about but it may be: https://github.com/mozilla/telemetry-dashboard/issues/625

These changes fix this by removing the version component from Line.getStateString and renaming it to Line.getColorId (this is the only use of that method).


chutten:
For submission date view, the separate colours per version is important so that dates where multiple versions are being used can be distinguished.

We could do this for build date view only, which would capture the most important cases and probably not be too confusing. Unfortunately it's unlikely that I'll get to it any time soon. (Firefox stuff keeping me busy)

I'll notify @hamilton here to see if the evolution view in GLAM (the Measurement Dashboard's upcoming successor) might benefit from fixing this before it launches.

chutten:
I do indeed believe that this is #625 

We might not wish to do this for the submission date view, as the same submission date can have multiple versions (the same x value can have multiple data points) for a single aggregate. See, for instance [this](https://mzl.la/32yyRbz).

How bad would it look in the code if we only turned to this mode for submission date view?

chutten (changes requested):
Let me know what you think about my earlier comment.

PR #636, opened 2019-12-04
redirect old probe dictionary to the new one at PTMO

spasovski:


chutten (approved):
Looks good to me

chutten:
Give it, say, fifteen minutes to update its way through the various caches and trigger the necessary scripts and then it should be live and testable.

PR #638, opened 2020-01-15
Bug 1605442 - mozaggregator missing prerelease data since 2019-01-09

acmiyaguchi:
Updating status as the service plays catch-up.

### DIFF
--- a/new-pipeline/dist.html
+++ b/new-pipeline/dist.html
@@ -32,7 +32,7 @@
             </form>
             <h1>Measurement Dashboard</h1>
             <div class="error-msg">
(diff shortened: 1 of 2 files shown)


akkomar (approved):

new-pipeline/dist.html
```diff
@@ -32,7 +32,7 @@
             </form>
             <h1>Measurement Dashboard</h1>
             <div class="error-msg">
-                <span style="font-weight: bold">Partial outage of pre-release data from 2019-12-18 to present. </span>
+                <span style="font-weight: bold">Partial outage of pre-release data from 2019-01-09 to present. </span>
```
robhudson, line 35: Just curious if that should be `2020-01-09`?

PR #641, opened 2020-04-13
Bug 1621979 - update the update orphaning dashboard.

tublitzed:
Making a few small updates to [this dashboard](https://telemetry.mozilla.org/update-orphaning/):

* Adjusting text/links around maintainers.
* Adding details for 72.0.2 watershed 

I'd originally come in here to update the `Firefox version is 72 or higher...` text, not realizing that actually the text is [dynamically generated](https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/update-orphaning/index.js#L819) and the response we're getting back is correct and up to date. We're considering FF versions that are 2 behind release to still be "up to date". Since the most recent FF version for the current dataset was 74 (should be updating [about now to 75](https://github.com/mozilla/telemetry-airflow/blob/master/dags/update_orphaning_dashboard_etl.py#L28-L29)) then the fact that we're still seeing 72 is expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ validation/
 html/data
 out.txt
 Histograms.json
+.DS_Store
--- a/update-orphaning/index.html
+++ b/update-orphaning/index.html
@@ -97,8 +97,11 @@ criteria are applied is the same order as they are listed above.</div>
 <div id="state-failure-code-startup-link" class="bar-chart-link"></div>
 </div><br/>
 <hr>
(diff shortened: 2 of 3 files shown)


tublitzed:
I'm on a new machine, just noticed that it's not configured to sign commits. Will sign, rebase, and re-open when ready.

tublitzed:
new PR is [here](https://github.com/mozilla/telemetry-dashboard/pull/642)

PR #649, opened 2021-01-25
Add Glean Dictionary to TMO

wlach:
Not sure if this is the best description, but I think it gets the job done.


chutten (approved):
Certainly descriptive enough for starters. This file's easy enough to edit when we come up with catchier tag lines.

PR #651, opened 2021-02-03
Remove arewesmoothyet.com reference

squarewave:
It is no longer maintained and the domain registration lapsed.


chutten (approved):

chutten:
Thanks for catching that!

PR #652, opened 2021-04-29
Update dictionary links

wlach:
* Use production URL for Glean Dictionary
* Make it a little more clear that the probe dictionary is for legacy telemetry


chutten (approved):
I have Feelings about the wording, but they're all because I'm trying to live in the FOG Future. These changes are an excellent choice for The Now.

PR #653, opened 2021-06-15
bug 1715050 - Remove Release-channel data display and Auth

chutten:
This returns us to a situation similar to the way it was before: without auth, without release aggregates as a choice.

See it live here: https://chutten.github.io/telemetry-dashboard/


badboy (approved):
:'(

fbertsch (approved):
Forward, to GLAM!

PR #655, opened 2021-07-21
Remove data fault notice, add GLAM button

chutten:
The data fault's been resolved.

GLAM is preferred if you have a moz LDAP, and the Dashboards are in maintenance mode (more-or-less), so replace the 'file a bug' button with 'use glam'.

Live preview: https://chutten.github.io/telemetry-dashboard/new-pipeline/dist.html


brizental (approved):
A lock is an interesting choice of icon here... But I guess that makes sense since GLAM is only for people who have an LDAP, right?

chutten:
> A lock is an interesting choice of icon here... But I guess that makes sense since GLAM is only for people who have an LDAP, right?

Correct. We use fontawesome's lock to annotate LDAP-gated dashboards elsewhere on the site, so it's become a bit of a convention.

PR #662, opened 2022-03-11
Use the AMO API v4

### LINKED ISSUE #661: Upgrade use of https://addons.mozilla.org/api/v3/ to /api/v4/
AMO api v3 is [deprecated](https://blog.mozilla.org/addons/2021/02/01/addons-mozilla-org-api-v3-deprecation/) - and was scheduled to be removed entirely at the end of 2021.  We're hoping to [finally remove](https://github.com/mozilla/addons/issues/1861) it very soon.  From a search on https://github.com/mozilla/ we've noted some use in this repo of the /api/v3 endpoint that will break once the api is deleted.  

In many cases just replacing `/api/v3/` with `/api/v4/` is sufficient but some fields/endpoints have been renamed/removed (there's a lengthy v4 [changelog](https://addons-server.readthedocs.io/en/latest/topics/api/overview.html#v4-api-changelog)).

If on the other hand those files/this repo is no longer in use/maintained please close the issue so we can mark it off our list,
thanks.

badboy:
v3 is deprecated and soon to be removed.

Fixes #661

---

The addon recommender is not linked anywhere from TMO directly, so maybe we should just remove it fully? Just changing it was the safer way though.

### DIFF
--- a/addon-recommender/js/main.js
+++ b/addon-recommender/js/main.js
@@ -12,7 +12,7 @@ const addonMappingURI =
 
 // The AMO API endpoint used to request the top addons.
 const topAddonsURI =
-  'https://addons.mozilla.org/api/v3/addons/search/?q=&app=firefox&type=extension&sort=users';
+  'https://addons.mozilla.org/api/v4/addons/search/?q=&app=firefox&type=extension&sort=users';
 
 function setupAutocomplete() {
   var suggestions = []


chutten (approved):
Agreed that this is the safest change to make.

As for removing it, well, I'd expect @Dexterp37 to know who to ask about TAAR's status

Dexterp37:
@chutten this can be safely removed without bothering the TAAR owners (which is Evgeny from MAML btw :) ).

I put this together with Roberto Vitillo to showcase the system, but we were the only consumers, so it can be safely removed :)

chutten:
> @chutten this can be safely removed without bothering the TAAR owners (which is Evgeny from MAML btw :) ).
> 
> I put this together with Roberto Vitillo to showcase the system, but we were the only consumers, so it can be safely removed :)

Alrighty, I'll file another issue for that work.

PR #665, opened 2022-09-08
redesigning

mcfunley:
I put ~1.5 hours into cleaning this up. 

- No longer showing tweets since those haven't moved since April, and it takes up a lot of space
- I added grafana and bq console
- I tried to organize them better
- Rewrote some descriptions slightly
- I dropped many links that weren't clearly working (open to discussion on these, I was quite aggressive I guess)

You can see it here: https://mcfunley.github.io/telemetry-dashboard/


chutten:
Thank you for the contribution, I'll have time to give this my full attention in the next couple days.

assets/dash.css
```diff
@@ -0,0 +1,60 @@
...
+    color: #aaa;
+  }
+
+  #header a.btn:hover {
+    color: #fff;
+  }
+
+
+  #logo {
+    height: 52px;
```
chutten, line 15: Visuals: The header height is a little heavy compared to other sites in the family (margin box height for GLAM is 48, Glean Dictionary 67 - this (coupled with h1 and padding) puts us at 104. With widescreen monitors, we can slim this down to focus more on the index content.

assets/dash.css
```diff
@@ -0,0 +1,60 @@
+#header {
+  background-color: #000 !important;
```
chutten, line 2: Curious: `!important` is a code smell from the days I last CSS'd. Which selector's more specific than #id?

assets/dash.css
```diff
@@ -0,0 +1,60 @@
...
+main .row {
+  margin-bottom: 1.75rem;
+}
+
+
+.tool {
+  display: flex !important;
+  align-items: flex-start !important;
+  border: 1px solid #fff;
+  box-sizing: border-box;
```
chutten, line 28: Nit: Bootstrap puts this on `*` so we can drop it (plus it only has effect if you're setting `width` or `height`)

index.html
```diff
@@ -7,160 +7,192 @@
     <title>Telemetry portal</title>
 
     <!-- Bootstrap and supporting libraries -->
-    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css">
-    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
-    <link rel="stylesheet" href="new-pipeline/style/dashboards.css">
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.1/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-iYQeCzEYFbKjA/T2uDLTpkwGzCiq6soy8tYaI1GyVh/UjpbCx/TYkiZhlZB6+fzT" crossorigin="anonymous">
```
chutten, line 10: Q: I'm not familiar with jsdelivr. What's to recommend it over the "official" bootstrapcdn?

index.html
```diff
@@ -7,160 +7,192 @@
...
+
+            <!--
+                 <a class="tool col" href="">
+                     <img src="" />
+                     <div>
+                         <h3></h3>
+                         <p></p>
+                     </div>
+                 </a>
+            -->
```
chutten, line 184: Checking in commented-out code should be avoided.

chutten (changes requested):
As we chatted about yesterday, there's a few things you're hitting upon with this.

1) What is the purpose of the TMO Index
 - Back when Georg was around I remember some consultations and data analysis about who used TMO and what for. No firm design was proposed at the time due to Thing 2
2) How does this relate to data.firefox.com or a new data.mozilla.com or otherwise `.com` Data Resources Index
  - TMO is `.org` which marks it a little more public than private, which has different requirements
  - But who knows about this, and who would care? Needs study and care
3) Is Public a category of things, or is it an attribute that some of our tools have? Is public default, or is private?
  - Historical context on this is a fun read (if we'd written it down)

As such I think that to get this landed we'll need to split this. The broader "What should TMO be" decision was and (I feel) remains something that needs consultation and design. This work can proceed via the usual means (Data Proposal Process, etc). The visual refresh could land with less delay.

To turn this visuals-only, we'll need the current list of tools (additions welcome) and the current public/SSO split (public is assumed to be default, private gets a little lock or some other mark). 

The Twitter embed doesn't even show when strict ETP is turned on (which I believe is Nightly's default), so I feel it can be omitted from the design (though mreid may have plans for improving its relevance, so we may need to find space for it someplace eventually)

Thoughts?

assets/dash.css
mcfunley: I agree re: smell but bootstrap is still full of `!important`. Not sure off the top of my head what was overriding this, but it was something `!important` somewhere else.

mcfunley:
I'd vote to merge it as-is and iterate - practically speaking I don't have a ton of time to revise and there aren't any showstoppers. The current TMO is full of broken stuff and getting incrementally better (rather than perfect) is my agenda.

@mreid-moz thoughts?

index.html
mcfunley: this is the bootstrap recommendation https://getbootstrap.com/docs/5.2/getting-started/download/#cdn-via-jsdelivr

Dexterp37:
All right! Thanks @mcfunley and @chutten . I will take this over now and push it to the finish line.

Dexterp37:
Closed in favour of #666 .

PR #674, opened 2023-05-18
Revert "Add error notice for DSRE-951 to Update Orphaning dash"

akkomar:
This reverts commit 618f7e3c1fca43dd3a026051f6993b5fadf3c1a0.

Dashboard has been fixed in https://github.com/mozilla/telemetry-airflow/pull/1703

### DIFF
--- a/update-orphaning/index.html
+++ b/update-orphaning/index.html
@@ -10,12 +10,6 @@
 <script type="text/javascript" src="index.js" charset="utf-8"></script>
 </head>
 <body onload="initDashboard()">
-  <div style="border: 2px solid #C13832; background-color: #D7D3C8; padding: 0.5em; color: #4D4E53; margin-bottom: 1em;">
-    <span style="font-weight: bold">Data Fault: Ongoing</span>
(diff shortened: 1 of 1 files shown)


update-orphaning/index.html
```diff
@@ -10,12 +10,6 @@
 <script type="text/javascript" src="index.js" charset="utf-8"></script>
 </head>
 <body onload="initDashboard()">
-  <div style="border: 2px solid #C13832; background-color: #D7D3C8; padding: 0.5em; color: #4D4E53; margin-bottom: 1em;">
```
kik-kik, line 13: The fix linked, does that apply for data from past dates? In this case maybe we should leave a note that this issue is not longer a problem from when the PR was merged?

kik-kik (approved):
Approved, but please check my comment before merging.

update-orphaning/index.html
akkomar: It should be fine - I have restarted couple of recent runs and folks are looking only at the latest aggregation here.

PR #679, opened 2024-11-14
Add the correct use-counters dahsboard

Dexterp37:


index.html
```diff
@@ -236,7 +236,7 @@ <h3>Dashboard Generator</h3>
                     </div>
                 </a>
 
-                <a class="tool col" href="https://mozilla.github.io/usecounters/index.html">
+                <a class="tool col" href="https://mozilla.cloud.looker.com/folders/1045">
```
chutten: We need to move this whole tile out of "Public Data and Tools" if it's in Looker

chutten (changes requested):

index.html
Dexterp37: You are indeed correct

chutten (approved):

PR #680, opened 2024-12-19
Add Experimenter Docs to the Data Portal

bochocki:
Add a link to Experimenter Docs in the Data Portal. I have trouble finding/remembering the link to Experimenter Docs sometimes, and the Data Portal seems like a great place for it to be.

### DIFF
--- a/index.html
+++ b/index.html
@@ -137,6 +137,14 @@
                     </div>
                 </a>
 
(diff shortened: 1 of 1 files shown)


chutten (approved):
One of these days we should probably work out an ordering scheme for this section, but for now this'll suit.

